PR #94, opened 2026-03-09
fix: support extras and source builds in pybun install

### LINKED ISSUE #93: Installation fails for packages requiring source builds or with extra markers
### Description
While using PyBun (v0.1.15) on `macos_arm64`, I encountered several issues during `pybun install`:

1. **404 error for extras**: `typer[all]` fails because it tries to fetch `https://pypi.org/pypi/typer[all]/json` which results in a 404.
2. **Missing sdist support**: Packages like `pydantic-core` (2.42.0) and `PyYAML` (6.0.3) fail to install because pre-built wheels are unavailable for this platform, and PyBun does not yet support installing from source distributions.

### Error Messages
- `pybun install: io error: http error: HTTP status client error (404 Not Found) for url (https://pypi.org/pypi/typer[all]/json)`
- `pybun install: The following packages have no pre-built wheel for your platform and require source builds (not yet supported): pydantic_core==2.42.0, PyYAML==6.0.3`

### Environment
- OS: macOS (Apple Silicon)
- PyBun version: 0.1.15

takurot:
## Description
- fix PyPI resolution so root requirements with extras (for example `typer[all]`) query the base package name and include extra-marked dependencies in the resolved graph
- add PyPI source distribution fallback during `pybun install` by downloading the selected sdist and installing it through the target Python environment when no compatible wheel exists
- extend resolver/lock output and tests to cover extras-aware dependency recording and sdist install behavior

## Motivation and Context
Issue #93 reported two install failures on `macos_arm64`:
- `typer[all]` queried `/pypi/typer[all]/json` and failed with 404
- packages without a platform wheel stopped at resolution even when an sdist was available

Fixes #93

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/Build changes

## How Has This Been Tested?
- [x] Unit tests added/updated
- [x] Integration tests added/updated
- [x] Manually tested on [OS/Platform]

Manual/automated checks run:
- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test`
- `cargo test --test '*'`
- `cargo audit`

## Checklist
- [x] My code follows the code style of this project (`cargo fmt`)
- [x] My code passes all lint checks (`cargo clippy`)
- [x] All tests pass (`cargo test`)
- [ ] I have updated the documentation accordingly
- [x] I have added tests to cover my changes (if applicable)
- [x] All new and existing tests pass

## Screenshots (if applicable)
- N/A

## Additional Notes
- `cargo audit` completed successfully but still reports existing allowed warnings for unmaintained crates (`async-std`, `instant`, `bincode`).

### DIFF
--- a/tests/e2e_general.rs
+++ b/tests/e2e_general.rs
@@ -1,5 +1,9 @@
 use assert_cmd::Command;
+use httpmock::prelude::*;
+use pybun::lockfile::Lockfile;
+use serde_json::json;
 use std::fs;
+use std::io::Write;
 use std::path::Path;
 use tempfile::tempdir;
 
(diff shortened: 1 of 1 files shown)


src/pypi.rs
```diff
@@ -376,56 +399,49 @@ impl PyPiClient {
...
-        deps: Vec<String>,
+        deps: ParsedDependencies,
         memory: &Arc<DashMap<String, Vec<CachedPackage>>>,
     ) -> Result<(), PyPiError> {
         if let Some(mut packages) = memory.get_mut(name)
             && let Some(pkg) = packages.iter_mut().find(|pkg| pkg.version == version)
         {
-            pkg.dependencies = Some(deps.clone());
+            pkg.dependencies = Some(deps.base.clone());
+            pkg.optional_dependencies = deps.optional.clone();
```
takurot: [P1] This update never runs for warm caches created before the extras work landed. `ensure_dependencies()` returns early once `cached_dependencies()` sees base deps, and legacy cache entries deserialize `optional_dependencies` to an empty map by default. That leaves `typer[all]` missing its extra deps until the cache is cleared. We need a schema/version bit (or similar) so legacy cache entries get refreshed once.

PR #128, opened 2026-04-01
PR-A2: enforce strict lock/hash verification

### LINKED ISSUE #115: feat: reproducibility hardening — strict lockfile verification and artifact provenance
## Problem

PyBun is already moving toward signed artifacts and reproducible installs, but the core install path is not yet strict enough for AI-agent and enterprise use:
- lock generation can still tolerate incomplete artifact integrity metadata on some paths
- verification is not enforced as a default expectation across `install`, `lock`, and `upgrade`
- agents and CI systems need to know that a lockfile is not only deterministic, but also cryptographically meaningful
- SBOM generation and vulnerability scanning help after the fact, but they do not guarantee that the resolved artifact set is itself trustworthy at install time

## Proposed Solution

Make artifact verification and provenance a first-class invariant of the lock/install flow.

### Scope
- Require a real `sha256` for every locked artifact; fail lock generation when hashes are missing
- Unify `install`, `lock`, and `upgrade` so all three produce and consume the same verification-grade artifact metadata
- Add a strict verification mode that is easy to adopt in CI and make it the default target behavior
- Record artifact provenance in the lockfile: source index URL, artifact URL, upload time, signer identity when available, and platform tag used for selection
- Emit structured JSON verification results so agents can distinguish hash mismatch, signature mismatch, provenance mismatch, and missing metadata
- Add drift diagnostics for lockfiles created by older PyBun versions

### CLI examples
```bash
pybun install --frozen --verify
pybun lock --verify
pybun upgrade --verify
```

### JSON output example
```json
{
  "status": "ok",
  "detail": {
    "verified": true,
    "artifacts": [
      {
        "package": "requests",
        "version": "2.32.0",
        "sha256": "...",
        "source_url": "https://files.pythonhosted.org/...",
        "index_url": "https://pypi.org/simple",
        "platform_tag": "py3-none-any"
      }
    ]
  }
}
```

## Why This Matters

If PyBun is going to be trusted as execution substrate for AI agents, it needs to guarantee more than fast resolution. It needs to guarantee that the exact artifact set an agent resolved is the artifact set that actually got installed, with a machine-verifiable chain of custody.

## Related
- `src/lockfile.rs`
- `src/downloader.rs`
- `src/security.rs`
- `src/commands.rs`
- `docs/PLAN.md` PR-A2

## Not a Duplicate Of
- #113: that issue is broader supply-chain security (`audit`, typosquatting, SBOM). This proposal is narrower and stricter: lockfile integrity and install-time verification semantics.
- #110: that issue is about experiment snapshots/import-based locks, not artifact verification guarantees for the standard install path.

takurot:
## Description

Enforce strict artifact hash verification across `install`, `lock`, and `upgrade` so placeholder hashes are rejected before lockfiles are written.

## Motivation and Context

Issue #115 requires reproducible installs to be cryptographically meaningful by default. This change makes missing hashes a hard failure, surfaces structured verification diagnostics in JSON, and warns when upgrading legacy lockfiles that still contain placeholder hashes.

Fixes #115

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/Build changes

## How Has This Been Tested?

- `cargo test`
- `cargo test --test '*'
- `just lint`
- `cargo audit`
- `cargo deny check licenses`
- `cargo build --release`
- `PATH=$(pwd)/target/release:$PATH python3 scripts/benchmark/bench.py -s run --format markdown`

- [x] Unit tests added/updated
- [x] Integration tests added/updated
- [x] Manually tested on [macOS]

## Checklist

- [x] My code follows the code style of this project (`cargo fmt`)
- [x] My code passes all lint checks (`cargo clippy`)
- [x] All tests pass (`cargo test`)
- [x] I have updated the documentation accordingly
- [x] I have added tests to cover my changes (if applicable)
- [x] All new and existing tests pass

## Screenshots (if applicable)

N/A

## Additional Notes

- Success JSON now includes `verified` and per-artifact verification metadata.
- Missing hashes return `E_VERIFY_MISSING_HASH` and downloader placeholder checks now fail with `MissingChecksum`.
- `upgrade` emits `W_LOCK_PLACEHOLDER_HASH` when older lockfiles contain placeholder hashes.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -1226,10 +1226,7 @@ async fn install(
         lock.add_package(Package {
             name: pkg.name.clone(),
             version: pkg.version.clone(),
-            source: PackageSource::Registry {
-                index: "pypi".into(),
-                url: "https://pypi.org/simple".into(),
-            },
+            source: registry_source_for_index(&source_index_url),
             wheel: selection.filename,
             hash: verified_hash,
             dependencies: pkg.dependencies.iter().map(ToString::to_string).collect(),
@@ -1423,13 +1420,15 @@ struct LockOutcome {
 
 fn is_missing_sha256(hash: Option<&str>) -> bool {
     match hash {
-        Some(value) => {
-            let normalized = value.trim();
-            normalized.is_empty()
-                || normalized == "placeholder"
-                || normalized == "sha256:placeholder"
-        }
+        Some(value) => crate::security::is_placeholder_hash(value),
         None => true,
+    }
+}
+
+fn registry_source_for_index(index_url: &str) -> PackageSource {
+    PackageSource::Registry {
+        index: "pypi".into(),
+        url: index_url.to_string(),
     }
 }
 
@@ -1653,10 +1652,7 @@ async fn lock_dependencies(args: &LockArgs, collector: &mut EventCollector) -> R
         lock.add_package(Package {
             name: pkg.name.clone(),
             version: pkg.version.clone(),
-            source: PackageSource::Registry {
-                index: "pypi".into(),
-                url: "https://pypi.org/simple".into(),
-            },
+            source: registry_source_for_index(&source_index_url),
             wheel: selection.filename,
             hash: verified_hash,
             dependencies: pkg.dependencies.iter().map(ToString::to_string).collect(),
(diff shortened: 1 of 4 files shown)


takurot:
Addressed in `affd052`.

- `upgrade` now writes consistent `PackageSource::Registry` metadata, and `install` / `lock` use the same source-construction helper.
- Placeholder-hash detection is centralized in `src/security.rs` and reused from both command verification and downloader verification.
- Added a regression test covering `upgrade --index <local-file>` lockfile source metadata.

Validated with:
- `cargo test --test cli_install --test cli_lock --test cli_upgrade --test json_output --test security_features --test pypi_integration`
- `cargo test --test json_schema`
- `cargo test --lib security::tests::placeholder_hashes_are_rejected`

PR #151, opened 2026-06-02
fix(run): propagate child Python process exit code to shell (#148)

### LINKED ISSUE #148: bug: pybun run does not propagate the exit code of the child python process
### Describe the bug
When running a Python script (or inline code) with `pybun run` or `pybun run --sandbox`, if the underlying Python process exits with a non-zero exit code (e.g. `1` due to an unhandled exception or explicit `sys.exit(1)`), the `pybun` CLI process still exits with code `0`.

### To Reproduce
Run the following command:
```bash
pybun run -c -- "import sys; sys.exit(42)"
echo "Exit code: $?"
```

**Expected behavior:**
The command should output the error and the CLI exit code should be `42`.

**Actual behavior:**
It prints `pybun run: inline code exited with code 42` but the shell exit code is `0`.

### Impact
This makes it difficult to detect failures in CI/CD pipelines or wrapper scripts, as automation tools rely on exit codes to determine the success/failure of commands.

### Proposed Solution
Propagate the exit code of the spawned python process to the `pybun` CLI wrapper (e.g., using `std::process::exit` in Rust or `sys.exit` in python shim wrappers).

takurot:
## Summary

Fixes #148.

`pybun run` was always exiting with code `0` even when the spawned Python process exited with a non-zero code. This broke CI/CD pipelines and wrapper scripts that rely on the exit code to detect failures.

**Root cause**: `RenderDetail` had no mechanism to carry a child-process exit code. `execute()` only called `std::process::exit(1)` on internal command errors, never on the script's own exit code.

**Fix**:
- Add `process_exit_code: Option<i32>` field to `RenderDetail`
- Add `with_process_exit_code(code: i32) -> Self` builder method
- In the `Commands::Run` match arm, attach the `RunOutcome.exit_code` to `RenderDetail` via `with_process_exit_code()`
- In `execute()`, after flushing output, call `std::process::exit(code)` when `process_exit_code` is `Some(non-zero)`

This correctly handles all run modes: script file, inline `-c`, and sandboxed execution.

## Behaviour change

| Scenario | Before | After |
|---|---|---|
| Script exits 0 | pybun exits 0 ✅ | pybun exits 0 ✅ |
| Script exits 42 (text mode) | pybun exits 0 ❌ | pybun exits 42 ✅ |
| Script exits 42 (json mode) | pybun exits 0 ❌ | pybun exits 42 ✅ |
| Sandbox blocks script (exit 1) | pybun exits 0 ❌ | pybun exits 1 ✅ |

JSON output (`--format=json`) is emitted **before** `std::process::exit`, so structured output is never lost.

## Test plan

- [x] `run_script_propagates_nonzero_exit_code` — text mode, script exit 42
- [x] `run_inline_code_propagates_nonzero_exit_code` — `-c` mode, exit 7
- [x] `run_script_exit_zero_still_succeeds` — exit 0 regression guard
- [x] `run_script_propagates_exit_code_json_mode` — JSON mode, exit 5, JSON still valid
- [x] `run_script_with_exit_code` — updated to assert `code(42)` (was `.success()`)
- [x] `sandbox` tests updated to assert `code(1)` where script fails due to policy
- [x] All 262+ existing tests pass
- [x] E2E tests (`cargo test --test e2e_general`) pass

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -315,6 +315,9 @@ pub async fn execute(cli: Cli) -> Result<()> {
                     cleanup,
                 }) => (
                     "x".to_string(),
+                    // TODO: propagate exit_code for `pybun x` the same way
+                    // `pybun run` does (with_process_exit_code). Tracked
+                    // separately to keep this PR scoped to issue #148.
                     RenderDetail::with_json(
                         summary,
                         json!({
@@ -704,13 +707,20 @@ pub async fn execute(cli: Cli) -> Result<()> {
         println!("{output}");
     }
 
-    // Exit with error code if command failed
+    // Flush stdout before any std::process::exit call. std::process::exit
+    // skips destructors, so a BufWriter around stdout (common on Windows)
+    // would otherwise silently discard buffered output.
+    let _ = std::io::Write::flush(&mut std::io::stdout());
+
+    // `is_error` and `process_exit_code` are mutually exclusive: the Err
+    // arm of every command sets is_error via RenderDetail::error() which
+    // leaves process_exit_code = None, while the Ok arm uses with_json()
+    // and may call with_process_exit_code(). is_error always takes priority.
     if is_error {
         std::process::exit(1);
     }
 
     // Propagate the child process exit code (e.g. from `pybun run`).
-    // Output has already been flushed above, so it is safe to exit now.
     if let Some(code) = process_exit_code
         && code != 0
     {
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -268,24 +268,23 @@ pub async fn execute(cli: Cli) -> Result<()> {
                             "audit": s.audit,
                         })
                     });
-                    (
-                        "run".to_string(),
-                        RenderDetail::with_json(
-                            summary,
-                            json!({
-                                "target": target,
-                                "exit_code": exit_code,
-                                "pep723_dependencies": pep723_deps,
-                                "pep723_backend": pep723_backend,
-                                "temp_env": temp_env,
-                                "cleanup": cleanup,
-                                "cache_hit": cache_hit,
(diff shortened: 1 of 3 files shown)


takurot:
## Code Review Findings

Reviewed by Rust specialist agent. Full test suite and clippy pass cleanly. No CRITICAL or HIGH issues found.

### MEDIUM — stdout not explicitly flushed before `std::process::exit`

**`src/commands.rs` (exit section)**

`println!` writes to a BufWriter-wrapped stdout. `std::process::exit` skips destructors, so the buffer may not flush — especially on Windows. The existing `is_error` exit path has the same latent bug; the new propagation path makes it observable.

**Fix**: add `let _ = std::io::Write::flush(&mut std::io::stdout());` before each `std::process::exit` call.

### MEDIUM — `Commands::X` exit code silently dropped

`pybun x` also spawns a subprocess with an `exit_code`, but the `Commands::X` arm never calls `.with_process_exit_code()`. Users running `pybun x ruff check . && echo ok` will always see exit 0 regardless of the tool's result. Should be tracked as a follow-up or fixed in this PR.

### LOW — `is_error` + `process_exit_code` interaction not documented

No invariant comment explains that these two flags are mutually exclusive. If a future change creates a `RenderDetail::error(...).with_process_exit_code(n)`, the child's code would be silently ignored (is_error wins due to ordering).

### LOW — `unwrap_or(-1)` for signal-killed child is ambiguous (pre-existing)

When a child is killed by a signal, `status.code()` returns `None`. Using `-1` wraps to 255 in the shell. Convention is `128 + signal_number`. Pre-existing issue, but new propagation makes it observable.

**Verdict**: Fix is functionally correct. Will address the two MEDIUM items.

PR #158, opened 2026-06-03
feat(sandbox): add default write restriction for system-critical paths (#150)

### LINKED ISSUE #150: feat: add file read/write access control in sandbox mode
### Describe the feature request
Currently, `--sandbox` isolates subprocess execution and network access by injecting `sitecustomize.py`. However, it does not restrict file access (such as writing to local files).

Many use cases for running untrusted AI-generated scripts in a sandbox require restricting where the script can read or write files (e.g., preventing writes to the system root, limiting read access to specific directories).

### Proposed Solution
Introduce options or environment variables to restrict file access:
- `--sandbox --allow-read <dirs>`
- `--sandbox --allow-write <dirs>`

Or implement a default behavior that blocks file writes outside the temporary environment or project workspace.

takurot:
## Summary

- When `--sandbox` is active **without** `--allow-write`, writes to system-critical paths are now blocked by default (`/etc`, `/usr`, `/bin`, `/sbin`, `/lib`, `/proc`, `/sys`, `/dev`, `/boot`; macOS adds `/System`, `/Library`, `/Applications`, `/private/etc`, `/private/var`)
- Writes to `/tmp`, Python venv prefix, and home directory remain allowed
- When `--allow-write` is explicitly set, existing behavior is preserved (allowlist restricts all other paths)
- JSON output includes a `default_deny_write` array showing the active default policy

## Changes

- `src/sandbox.rs`: new `default_system_deny_write_paths()` returns platform-specific system paths; `apply_python_sandbox()` passes `PYBUN_SANDBOX_DEFAULT_DENY_WRITE` env var (non-empty only when `allow_write` is empty); sitecustomize.py gains `_DEFAULT_DENY_WRITE` / `_HAS_DEFAULT_DENY_WRITE` / `_is_in_denied()` and activates `_patch_filesystem()` for default deny too; audit correctly increments `blocked_file_writes` when sandbox intercepts before OS
- `src/commands.rs`: `SandboxInfo` struct gains `default_deny_write: Vec<String>`; JSON `sandbox` object includes `default_deny_write`

## Test plan

- [x] `sandbox_default_blocks_write_to_etc` — `/etc` writes blocked by default sandbox
- [x] `sandbox_default_allows_write_to_tmp` — `/tmp` writes remain allowed
- [x] `sandbox_default_write_restriction_audit_counts_blocked_writes` — audit shows `blocked_file_writes:2` (sandbox intercepts before OS; discriminating test)
- [x] `sandbox_explicit_allow_write_overrides_default_restriction` — `--allow-write` still works as before
- [x] `sandbox_json_output_includes_default_deny_write_paths` — JSON includes `default_deny_write` field
- [x] All 15 sandbox tests pass
- [x] All E2E tests pass (`cargo test --test e2e_general`)

Closes #150


takurot:
## Code Review Response

A subagent reviewed the PR and found the following issues. All have been addressed in the follow-up commit `a7e1da6`.

### CRITICAL (Fixed)

**1. `_is_in_denied` fail-open on exception** — The `except Exception` clause returned `False`, which silently granted write access when path evaluation failed. Changed to `return True` (fail-closed) so any unexpected exception blocks the write. Security guards should never fail-open.

**2. Empty-string path bypass** — `os.path.realpath(os.path.abspath(""))` returns the CWD, which could bypass the deny check. Added a `if not path: return _orig_open(...)` guard early in `_checked_open`.

### HIGH (Fixed)

**3. macOS venv under `/private/var/folders` incorrectly blocked** — The default deny list includes `/private/var`, but macOS PEP 723 isolated venvs and user caches often live under `/private/var/folders/...`. The fix: in the default-deny branch, additionally check `_is_allowed(path, [])` which exempts sys prefixes (already populated in `_ALWAYS_ALLOW_READ`) before blocking.

**4. Duplicated `default_deny_write` logic in `commands.rs`** — The `if args.allow_write.is_empty() { ... }` condition was copy-pasted in two places. Fixed by adding `default_deny_write: Vec<String>` to `SandboxGuard` and reading from `guard.default_deny_write.clone()` at both call sites. Single source of truth.

### LOW (Fixed)

**5. `_write_audit` file handle leak** — Changed to `with _orig_open(...) as _f:` to ensure proper cleanup even under PyPy or late interpreter shutdown.

### Tests Added

- `sandbox_explicit_allow_write_yields_empty_default_deny_write_in_json` — verifies `default_deny_write: []` when `--allow-write` is specified
- Strengthened `sandbox_json_output_includes_default_deny_write_paths` to assert non-empty content (checks for `/etc` or `/usr`)

### Not Fixed (with rationale)

- **Symlink bypass test**: `_normalize_path` calls `os.path.realpath()` which already follows symlinks. Existing path normalization handles this case. A dedicated integration test would require creating symlinks pointing into denied paths, which is complex and platform-dependent. Tracked as future work.
- **Python-level unit tests for `SITECUSTOMIZE_PY`**: The policy logic is tested adequately via integration tests. Adding Python-level unit tests would require extracting the sitecustomize into a separate file, which is a larger refactor.
- **`default_system_deny_write_paths` allocation**: The reviewer suggested `&'static [&'static str]`, but the function uses conditional compilation (`#[cfg(target_os = "macos")]`) which requires runtime construction. The allocation is call-site bounded (called once per sandbox session) and not performance-critical.

PR #160, opened 2026-06-03
fix(security): pin rustls-webpki >=0.103.13 to clear 4 RustSec CVEs (Issue #156)

### LINKED ISSUE #156: security: cargo audit reports rustls-webpki RustSec advisories
## Summary

`cargo audit` currently fails on `main` due to RustSec advisories in the `rustls-webpki` dependency tree.

This was found during the comprehensive evaluation run on 2026-06-03 against local `main` at `e5b005bb5dee915152773483027dd06dc7023f30` (`pybun` 0.1.16). I checked for duplicates with `RUSTSEC-2026-0104`, `rustls-webpki`, and `cargo audit`; no existing issue matched this concrete dependency failure. This is separate from #113, which is a broader product feature request for PyBun package supply-chain scanning.

## Reproduction

```bash
cargo audit
```

Observed result:

```text
error: 4 vulnerabilities found!
warning: 4 allowed warnings found
```

The failing advisories are all for `rustls-webpki 0.103.8`:

- `RUSTSEC-2026-0104`: Reachable panic in certificate revocation list parsing
- `RUSTSEC-2026-0098`: Name constraints for URI names were incorrectly accepted
- `RUSTSEC-2026-0099`: Name constraints were accepted for certificates asserting a wildcard name
- `RUSTSEC-2026-0049`: CRLs not considered authoritative by Distribution Point due to faulty matching logic

Dependency path from `cargo audit`:

```text
rustls-webpki 0.103.8
└── rustls 0.23.35
    ├── tokio-rustls 0.26.4
    │   ├── reqwest 0.12.25
    │   │   └── pybun 0.1.16
    │   └── hyper-rustls 0.27.7
    │       └── reqwest 0.12.25
    ├── reqwest 0.12.25
    ├── quinn-proto 0.11.14
    │   └── quinn 0.11.9
    │       └── reqwest 0.12.25
    ├── quinn 0.11.9
    └── hyper-rustls 0.27.7
```

`cargo audit` reports the solution as upgrading `rustls-webpki` to patched ranges such as `>=0.103.13` for `RUSTSEC-2026-0104` and `>=0.103.12` / `>=0.103.10` for the others.

## Additional warnings

The same run also reports allowed warnings for:

- `RUSTSEC-2025-0052`: `async-std` unmaintained, via `httpmock`
- `RUSTSEC-2025-0141`: `bincode` unmaintained
- `RUSTSEC-2024-0384`: `instant` unmaintained, via `notify-types`
- `RUSTSEC-2026-0097`: `rand` unsound with a custom logger using `rand::rng()`, via `quinn-proto`

## Expected behavior

The security audit gate should pass, or any intentional exceptions should be documented in policy with rationale. Given the public TLS parsing advisories, the preferred fix is to update the dependency tree so `rustls-webpki` resolves to a patched version.

takurot:
## Summary

Closes #156.

`cargo audit` was failing on `main` due to four CVEs in `rustls-webpki 0.103.8`:

| Advisory | Summary | Patched in |
|---|---|---|
| RUSTSEC-2026-0104 | CRL parsing reachable panic | >=0.103.13 |
| RUSTSEC-2026-0098 | URI name constraint incorrectly accepted | >=0.103.12 |
| RUSTSEC-2026-0099 | Wildcard name constraint bypass | >=0.103.10 |
| RUSTSEC-2026-0049 | CRL authority matching fault | >=0.103.10 |

`Cargo.lock` already resolved to `rustls-webpki 0.103.13` (which clears all four advisories). This PR adds two explicit safeguards to prevent silent regression:

1. **`Cargo.toml`**: Added `rustls-webpki = { version = ">=0.103.13", optional = true }` — the Cargo resolver now enforces the patched floor at resolve time.
2. **Regression test**: `rustls_webpki_version_is_at_least_patched_floor` in `tests/security_features.rs` parses `Cargo.lock` at test-time and hard-fails CI if the version regresses.

The four pre-existing allowed warnings (`RUSTSEC-2024-0384`, `RUSTSEC-2025-0052`, `RUSTSEC-2025-0141`, `RUSTSEC-2026-0097`) remain silenced in `.cargo/audit.toml` with upstream-tracking rationale — they are not affected by this change.

## Test plan

- [x] `cargo audit` exits 0
- [x] `cargo clippy --all-targets --all-features -- -D warnings` exits 0
- [x] `cargo test` — all tests pass (no regressions)
- [x] New test `rustls_webpki_version_is_at_least_patched_floor` passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1937,6 +1937,7 @@ dependencies = [
  "predicates",
  "regex",
  "reqwest",
+ "rustls-webpki",
  "semver",
  "serde",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,8 @@ tokio = { version = "1.48.0", features = ["full"] }
 reqwest = { version = "0.12.25", default-features = false, features = ["blocking", "json", "rustls-tls", "stream"] }
 # Floor pin: reqwest pulls in rustls → rustls-webpki transitively.
 # >= 0.103.13 clears RUSTSEC-2026-0104/0098/0099/0049 (Issue #156).
-rustls-webpki = { version = ">=0.103.13", optional = true }
+# Non-optional so the version floor is compile-time enforced in every build.
+rustls-webpki = ">=0.103.13"
 futures = "0.3.31"
 dashmap = "6.1"
 mimalloc = { version = "0.1.43", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,9 @@ fs2 = "0.4"
 notify = { version = "7.0", optional = true }
 tokio = { version = "1.48.0", features = ["full"] }
 reqwest = { version = "0.12.25", default-features = false, features = ["blocking", "json", "rustls-tls", "stream"] }
+# Floor pin: reqwest pulls in rustls → rustls-webpki transitively.
+# >= 0.103.13 clears RUSTSEC-2026-0104/0098/0099/0049 (Issue #156).
+rustls-webpki = { version = ">=0.103.13", optional = true }
 futures = "0.3.31"
 dashmap = "6.1"
 mimalloc = { version = "0.1.43", optional = true }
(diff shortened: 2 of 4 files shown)


takurot:
## Code Review Findings (subagent review)

### [HIGH] `optional = true` — floor pin is not compile-time enforced

**`Cargo.toml` line 38**

```toml
# Current (problematic)
rustls-webpki = { version = ">=0.103.13", optional = true }
```

`optional = true` without an activating feature means the crate is never compiled into any build. Cargo still enforces the version constraint at lock-file generation time (which is why `cargo audit` passes today), but the floor provides no protection if `Cargo.lock` is regenerated by `cargo update`. A plain non-optional dep makes the constraint genuinely load-bearing.

**Fix:** Remove `optional = true`.

---

### [MEDIUM] Test only checks the first `rustls-webpki` block

If the dep tree ever resolves two simultaneous versions, only the first one is validated. The second (potentially vulnerable) version would silently pass.

**Fix:** Collect all versions, assert all are >= 0.103.13.

---

### [MEDIUM] Hand-rolled semver parsing — use the `semver` crate instead

`split('.')` panics on pre-release tags (e.g. `0.103.0-alpha.1`). The `semver` crate is already a direct dependency of pybun.

**Fix:** Use `semver::Version::parse()` for comparison.

PR #162, opened 2026-06-04
fix(resolver): correct ABI resolution — select Python-version-matched wheels (Issue #161)

### LINKED ISSUE #161: [Bug] Incorrect ABI resolution on macOS 26.x / Python 3.11 resolves Python 3.10 wheels (cpython-310-darwin.so)
### Describe the bug
When running `pybun install` on macOS 26.5.1 under a Python 3.11.15 virtual environment (`.venv`), PyBun resolves and installs Python 3.10 wheels instead of Python 3.11 wheels.

For example, when resolving `pyarrow>=14.0`, PyBun installs the Python 3.10 wheel containing `cpython-310-darwin.so` extensions. This results in the following runtime import error under Python 3.11:
`ModuleNotFoundError: No module named 'pyarrow.lib'`

### Environment
- **OS**: macOS 26.5.1 (Apple Silicon)
- **Python**: 3.11.15 (inside `.venv` managed by Python's built-in `venv` module)
- **PyBun CLI Version**: 0.1.18 (channel: stable)

### Steps to Reproduce
1. Prepare a project directory with `pyproject.toml`:
   ```toml
   [project]
   name = "test-project"
   version = "0.1.0"
   dependencies = [
       "pyarrow>=14.0",
   ]
   ```
2. Create and activate a Python 3.11 virtual environment:
   ```bash
   python3.11 -m venv .venv
   source .venv/bin/activate
   ```
3. Run `pybun install` (without any prior lockfiles):
   ```bash
   pybun install
   ```
4. Verify the installed shared library extension suffix for `pyarrow` in the `.venv` directory:
   ```bash
   ls -la .venv/lib/python3.11/site-packages/pyarrow/*darwin.so
   ```

### Actual Behavior
The installed extensions are tagged with Python 3.10 (e.g., `_acero.cpython-310-darwin.so`), and trying to run `python -c "import pyarrow"` throws `ModuleNotFoundError: No module named 'pyarrow.lib'` because Python 3.11 cannot locate Python 3.11 extensions.

### Expected Behavior
PyBun should correctly identify the host Python version (Python 3.11) and resolve/install Python 3.11 wheels (e.g., `cp311` tags containing `cpython-311-darwin.so`).

### Additional Context
- Deleting the `pybun.lockb` lockfile and performing a fresh install resolves the same Python 3.10 wheel, which suggests a bug in the resolver or platform target detection logic under macOS 26.x environment.
- Temporarily bypassing with `pip install --force-reinstall pyarrow` inside the activated virtual environment successfully downloads and overwrites with the correct `cp311` wheel.

takurot:
## Summary

- **Root cause**: `select_artifact_for_platform` ranked wheels by platform tag only (e.g., `macosx_14_0_arm64`) and ignored the Python interpreter tag (`cp310`, `cp311`, …). When both cp310 and cp311 wheels were available, the resolver could pick the wrong version, leading to `ModuleNotFoundError` at import time (e.g., `pyarrow.lib` missing when installed into a Python 3.11 venv).
- **Fix**: Parse Python/ABI tags from wheel filenames, filter incompatible wheels, and rank exact-version matches highest.
- **Backward compat**: Wheels with unknown/missing tags (legacy index fixtures) are treated as compatible — no existing behavior broken.

### `src/resolver.rs`
- `Wheel` struct: add `python_tag: Option<String>` and `abi_tag: Option<String>`
- `parse_wheel_tags(filename)`: extract Python/ABI tags using PEP 427 position rules (3rd and 2nd components from the right of the stem)
- `python_version_to_cp_tag(version)`: convert `"3.11"` → `"cp311"`
- `is_wheel_python_compatible(python_tag, abi_tag, active_cp)`: PEP 425 rules — exact match, abi3 stable ABI (active >= minimum), pure-Python (`py3`)
- `select_artifact_for_platform_with_cp(pkg, platform_tags, active_cp)`: filters to only compatible wheels, then ranks by Python version match score
- `select_artifact_for_platform`: delegates to `_with_cp` using `active_python_cp_tag()` (auto-detected, cached)
- `rank_wheel`: +30 pts for exact Python match, +15 for abi3/unknown, +10 for py3

### `src/index.rs`, `src/pypi.rs`
- Derive `python_tag`/`abi_tag` via `parse_wheel_tags` when constructing `Wheel` from JSON index fixtures or PyPI cache

## Test plan

- [x] 20 new unit tests in `src/resolver.rs`:
  - `parse_wheel_tags_*` (6 tests): filename parsing for cp310, cp311, py3, abi3, hyphenated names, URL paths
  - `python_version_to_cp_tag_*` (3 tests): version string → cp tag conversion
  - `is_wheel_python_compatible_*` (6 tests): exact match, version mismatch, py3, abi3 boundary conditions
  - `select_artifact_*` (5 tests): cp311 preferred over cp310, cp310 preferred over cp311, incompatible-only falls back to sdist, abi3 fallback, py3 always compatible
- [x] All existing tests pass (`cargo test` — 282+ tests across all suites)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — 0 errors

Closes #161

### DIFF
--- a/docs/PLAN.md
+++ b/docs/PLAN.md
@@ -83,6 +83,11 @@
   - Current: `init_project()` に `&mut EventCollector` を追加し、stdin が TTY でない場合は `E_INIT_NOT_INTERACTIVE` diagnostic（`suggestion` フィールドに `pybun init --yes` 案内）を push してから早期 return。テキストモードの stderr にも `--yes` を含むメッセージを出力。
   - Tests: `tests/cli_init.rs` に3件追加 — `init_non_tty_without_yes_fails_with_hint`（テキスト出力に --yes 含む）、`init_non_tty_without_yes_json_fails_with_hint`（JSON diagnostics に suggestion フィールド含む）、`init_non_tty_with_yes_succeeds`（--yes 指定時は非 TTY でも成功）。
 
+- [DONE] PR-BUG161: ABI resolution fix — correct Python version wheel selection (Issue #161)
+  - Goal: `pybun install` が Python 3.11 環境で cp310 wheel を選択してしまうバグを修正する。
+  - Current: `Wheel` struct に `python_tag: Option<String>` / `abi_tag: Option<String>` フィールドを追加。`parse_wheel_tags()` がホイールファイル名から Python/ABI タグを解析（位置は末尾から -3, -2）。`python_version_to_cp_tag()` が "3.11" → "cp311" 変換。`is_wheel_python_compatible()` が PEP 425 互換ルール（exact match / abi3 stable ABI / py3 pure-Python）を判定。`select_artifact_for_platform_with_cp()` が互換 wheel だけをフィルタリングして best-rank を選択。`select_artifact_for_platform()` は auto-detect CPython タグで同関数に委譲。`index.rs` / `pypi.rs` の `Wheel` 構築箇所も `parse_wheel_tags()` でタグを付与。
+  - Tests: `src/resolver.rs` に20件追加 — `parse_wheel_tags_*`（6件）、`python_version_to_cp_tag_*`（3件）、`is_wheel_python_compatible_*`（6件）、`select_artifact_*`（5件）。全382テストパス。`cargo clippy --all-targets --all-features -- -D warnings` 0エラー。`cargo fmt -- --check` 差分なし。
+
 ### P2 (Post-GA Improvement)
 - PR-A8: Runtime catalog hardening（実チェックサム管理 + 3.13 preview）
   - Goal: 埋め込み runtime metadata の完全性・更新性を強化し、3.13 preview を段階導入。
--- a/src/index.rs
+++ b/src/index.rs
@@ -60,15 +60,20 @@ fn build_index(packages: Vec<IndexPackage>) -> InMemoryIndex {
             let wheels = pkg
                 .wheels
                 .iter()
-                .map(|w| Wheel {
-                    file: w.file.clone(),
-                    url: None,
-                    hash: w.hash.clone(),
-                    platforms: if w.platforms.is_empty() {
-                        vec!["any".into()]
-                    } else {
-                        w.platforms.clone()
-                    },
+                .map(|w| {
+                    let (python_tag, abi_tag) = crate::resolver::parse_wheel_tags(&w.file);
+                    Wheel {
+                        file: w.file.clone(),
+                        url: None,
+                        hash: w.hash.clone(),
+                        platforms: if w.platforms.is_empty() {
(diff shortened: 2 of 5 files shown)


takurot:
## Code Review Response

A thorough code review was conducted by a Rust specialist. Here are the findings and how they were addressed:

### HIGH issues (fixed in commit 1233e7b)

**1. `cp_tag_ge` — duplicated match arms and future cliff for Python 10.x**
- The `match digits.len()` had identical `2 =>` and `3 =>` arms (dead code), and a `_ => None` that would break for Python 10.x (4-digit tags like `cp100`).
- **Fix**: Replaced with `split_at(1)` — major is always exactly 1 digit, minor is the remainder. Handles any version length correctly.

**2. Compressed python tags (e.g., `cp310.cp311`) silently rejected**
- PEP 425 allows multi-version compressed tags in wheel filenames. The previous code treated `"cp310.cp311"` as a literal tag, causing it to never match.
- **Fix**: `is_wheel_python_compatible` now splits `ptag` on `'.'` and checks each component. For abi3 wheels, compatibility is determined by the oldest component. For CPython-specific wheels, any matching component is sufficient.

### MEDIUM issues (fixed in commit 1233e7b)

**3. `parse_wheel_tags` — empty tags from malformed double-dash filenames**
- A filename like `a--1.0--cp311--linux.whl` would produce `Some("")` rather than `None`.
- **Fix**: Added an early return of `(None, None)` if either extracted tag is empty.

**4. `build_resolved` in pypi.rs — guard only checked `python_tag.is_some()`**
- If a cached entry had `python_tag = Some(...)` but `abi_tag = None` (theoretically possible with a hand-crafted cache), it would skip re-parsing.
- **Fix**: Guard now requires both `python_tag.is_some() && abi_tag.is_some()` before skipping `parse_wheel_tags`.

**5. `rank_wheel` `_ => {}` arm — added clarifying comment**
- The arm now explains it covers non-CPython interpreters (e.g., PyPy `pp311`) that passed compatibility checking but have no specific score bonus.

### LOW issues (documented, no code change needed)

**6. `active_python_cp_tag` uses `get_python_version()` (PATH-based detection)**
- This is a pre-existing issue: if the active venv Python differs from `python3` on PATH, the wrong CP tag may be cached. All new tests use `select_artifact_for_platform_with_cp()` directly to bypass the cache — good design. Will track as a follow-up issue.

**7. Test fixture `source-only-0.5.0-win_amd64.whl` uses non-standard filename**
- The 4-component name parses to `(None, None)` and is correctly excluded by platform tag filtering (not Python tag filtering). Low priority, tracked for future fixture cleanup.

PR #164, opened 2026-06-05
fix(build): emit actionable hint when build package is missing (Issue #134)

takurot:
Closes #134

## Summary

- Detects `No module named build` in `python -m build` stderr
- Pushes a structured diagnostic (`E_BUILD_MISSING_BUILD_PKG`) with `suggestion: "pybun add build --dev\n  or: pip install build"`
- In text mode, prints an explicit `hint:` line to stderr so users see the fix immediately
- Applies `pre_diag_count` guard at the `Commands::Build` call site so the generic fallback error is suppressed when a structured diagnostic was already recorded (same pattern as `Commands::Init`)

## Test plan

- [x] `build_no_module_build_gives_hint_in_json` — JSON output contains diagnostic with code `E_BUILD_MISSING_BUILD_PKG` and suggestion containing `pybun add build`
- [x] `build_no_module_build_gives_hint_in_text` — text mode stderr contains `pybun add build` or `pip install build`
- [x] Existing `cli_build` tests unchanged and passing (5/5)
- [x] Full test suite: 0 failures across all test crates
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt` applied

### DIFF
--- a/docs/PROMPT.md
+++ b/docs/PROMPT.md
@@ -78,3 +78,91 @@ gh run view    # 詳細確認
 - 依存関係のあるPRがマージされていることを確認
 - `--format=json` オプションは全コマンドで対応すること（AI-friendly）
 - macOS/Linux を優先、Windows はスタブで対応
+
+---
+
+# Issue 実装プロンプト
+
+@docs/PROMPT.md に従って Issue #<番号> を実装する。
+
+## 実装フロー（Issue番号指定）
+
+### 1. Issue 内容の確認
+```bash
+gh issue view <番号>
+```
+- Issue の要件・背景・受け入れ条件を把握する
+- 関連する既存コード・テスト・ドキュメントを調査する
+
+### 2. ブランチ作成
+- `feature/issue-<番号>-<簡潔な説明>` の形式でブランチを作成
+  - 例: `feature/issue-157-benchmark-hermetic`
+- `main` ブランチから分岐すること
+
+### 3. 適切な Skills の選択と TDD 実装
+- タスクに応じて以下の Skills を選択して使うこと:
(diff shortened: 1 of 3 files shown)


takurot:
## Code Review — Issue #134 fix

### [HIGH] Detection substring should anchor to CPython's quoted format

CPython 3.x actually outputs:
```
/path/to/python3: No module named 'build'
```
(with single quotes around the module name). The current check `stderr.contains("No module named build")` matches, but anchoring to the quoted form reduces false-positive risk and is more precise.

**Fix:** Check both forms for safety: `"No module named 'build'"` and `"No module named build"`.

---

### [HIGH] New tests must be gated `#[cfg(unix)]`

`setup_no_module_build_project` writes a `#!/bin/sh` script. On Windows `CreateProcess` cannot exec a shebang file — both new tests will fail to spawn and error out rather than skip. Existing Windows CI (preview) would break.

**Fix:** Add `#[cfg(unix)]` to `build_no_module_build_gives_hint_in_json` and `build_no_module_build_gives_hint_in_text`.

---

### [MEDIUM] `pre_diag_count` guard needs a comment

The guard is correct and matches the `Commands::Init` pattern, but is invisible to future maintainers. A brief comment prevents the guard from being removed as "dead code".

---

### [LOW] Unnecessary `let suggestion` allocation

`with_suggestion` accepts `impl Into<String>` — the intermediate binding can be removed.

### [LOW] JSON test does not assert `code == "E_BUILD_MISSING_BUILD_PKG"`

The public diagnostic code field is untested; a rename would go undetected.

---

**Verdict: WARNING** — Directionally correct, no CRITICAL issues. The two HIGH items (detection anchor + `#[cfg(unix)]` gate) must be addressed before merge.

PR #170, opened 2026-06-07
fix(cli): JSON help output and consistent diagnostics (Issue #126)

### LINKED ISSUE #126: enhancement: PyBun CLI Usability and Diagnostic Improvements
### Description
During functional evaluation, several usability and diagnostic gaps were identified:

### Gaps
1. **`pybun build` Dependency Management**:
   - Currently, `pybun build` fails if the `build` package is not pre-installed in the venv.
   - **Fix**: PyBun should ideally bundle or automatically install build dependencies (similar to `pip install build`).

2. **Help output in JSON**:
   - `pybun --help --format=json` does not currently output JSON.
   - **Fix**: To be truly "Agent-First," all help information should be available in a machine-readable format.

3. **Inconsistent Diagnostics in JSON**:
   - Some error messages (like `pyproject.toml` parsing errors) are not consistently reported as `Diagnostic` objects in the JSON envelope's `diagnostics` array.

### Suggested Action
Include these improvements in upcoming maintenance releases to improve the developer and agent experience.

takurot:
Closes #126

## Summary
- Adds machine-readable `--help --format=json` support: clap intercepts `--help`/`-h` before our normal command dispatch runs (printing plain text), so `main.rs` now scans raw argv for the `--format=json` + `--help`/`-h` combination *before* `Cli::parse()` and, when matched, renders a JSON envelope (`version`/`command`/`status`/`detail`) describing the command (and one level of subcommand) via clap's introspection API (`get_subcommands`, `get_arguments`, `render_usage`, etc).
- Fixes inconsistent diagnostics reporting: `Commands::Install` and `Commands::Lock` previously built the error `RenderDetail` directly without pushing a `Diagnostic` into the JSON envelope's `diagnostics` array (unlike every other command arm), so generic errors (e.g. missing `pyproject.toml`, missing script) were silently absent from `diagnostics`. Both arms now use the `pre_diag_count` guard pattern already established for `run_build`/`Commands::Init`, so a fallback `Diagnostic::error` is pushed only when the underlying function didn't already record a structured one (avoiding double diagnostics for resolve errors).
- `pybun build` auto-installing the `build` package (gap 1 in the issue) is left out of scope — Issue #134 already added an actionable `E_BUILD_MISSING_BUILD_PKG` hint pointing users to `pybun add build --dev` / `pip install build`.

## Test plan
- [x] `src/cli.rs::help_tests` — unit tests for `scan_help_request` (top-level / subcommand / split `--format json` / text-format-is-ignored) and `json_help_envelope` envelope construction
- [x] `tests/cli_smoke.rs::help_supports_json_format` — `pybun --help --format=json` returns a valid envelope with `command == "pybun --help"` and `detail.subcommands` containing `install`
- [x] `tests/cli_smoke.rs::subcommand_help_supports_json_format` — `pybun install --format json --help` returns `detail.args` containing `offline`
- [x] `tests/cli_smoke.rs::text_format_help_is_unaffected` — regression guard that plain `--help` still prints text, not JSON
- [x] `tests/cli_install.rs::install_json_output_reports_error_in_diagnostics_array` — generic install error now appears as an error-level `Diagnostic`
- [x] `tests/cli_lock.rs::lock_json_output_reports_error_in_diagnostics_array` — generic lock error (missing script) now appears as an error-level `Diagnostic`
- [x] `cargo test` — all 288+ unit/integration tests pass
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — 0 warnings


takurot:
## Self-review findings (addressed)

During implementation review I identified two HIGH-severity issues, both now fixed in 8ae6733:

### [HIGH] `scan_help_request` misparsed value-taking long options as path segments
**File:** `src/cli.rs`

Only `--format` and `--progress` were special-cased as taking a following value. Any other value-taking long option (`--script`, `--require`, `--lock`, `--index`, etc.) fell through to the catch-all `path.push(...)` arm — e.g. `pybun lock --script foo.py --format=json --help` produced `command: "pybun lock foo.py --help"` instead of `"pybun lock --help"`.

**Fix:** Replaced the raw-argv path reconstruction with a walk of the `clap::Command` tree alongside the scan. A token only extends the path when it actually names a subcommand of the command reached so far — option values that don't coincide with a subcommand name are simply ignored. This avoids needing an exhaustive list of value-taking flags.

Added regression tests `ignores_option_values_that_resemble_subcommand_names` and `renders_help_envelope_with_correct_command_name_when_option_value_looks_like_a_subcommand`.

### [HIGH] `pre_diag_count` guard counted all diagnostics, not just errors
**Files:** `src/commands.rs`, `src/schema.rs`

The `Install`/`Lock`/`Build`/`Init` dispatch arms guarded the generic fallback error with `collector.diagnostic_count() == pre_diag_count`. Since `diagnostic_count()` counts diagnostics at every level (error/warning/info/hint), any info or warning diagnostic emitted before a failure (e.g. progress messages) would make the count differ from the baseline and incorrectly suppress the fallback error — leaving a failed command with `status: "error"` but no error-level diagnostic explaining why.

**Fix:** Added `EventCollector::error_diagnostic_count()` (counts only `DiagnosticLevel::Error` entries) and switched all four guard sites from `pre_diag_count`/`diagnostic_count()` to `pre_error_count`/`error_diagnostic_count()`.

---
All tests pass (`cargo test`), `cargo clippy --all-targets --all-features -- -D warnings` is clean, and `cargo fmt -- --check` is clean.

PR #179, opened 2026-06-09
perf(module-find): eliminate stat syscalls and add parallel subdir scanning (Issue #135)

takurot:
## Summary

- Replaces `path.is_dir()`/`path.is_file()` with `DirEntry::file_type()` in the scan inner loop — the OS already returns file type from `readdir()`, so each entry previously paid one extra `stat()` syscall for nothing
- When a directory has more than `PARALLEL_SUBDIR_THRESHOLD` (10) immediate subdirectories, scoped threads are used to process them concurrently, amortizing traversal cost for large trees like the Python stdlib (~2300+ modules)
- Adds `ScanResult{modules, duration_us}`, `scan_directory_timed()`, and `parallel_scan_timed()` so scan timing is measurable without a separate Instant at call sites
- Exposes `duration_us` in `--scan` JSON `detail` output and `--benchmark` text output

## Test plan

- [x] 4 new unit tests in `src/module_finder.rs`: `test_scan_directory_timed_returns_duration`, `test_parallel_scan_timed_returns_duration`, `test_scan_with_many_subdirs_finds_all_modules` (15 packages, verifies parallel path), `test_scan_with_few_subdirs_uses_sequential_path`
- [x] 3 new E2E tests in `tests/module_finder.rs`: `test_scan_json_includes_duration_us`, `test_scan_parallel_finds_all_modules_in_large_structure`, `test_scan_benchmark_includes_duration_in_text_output`
- [x] All existing tests pass (`cargo test`)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — zero errors

Closes #135


takurot:
## Code Review Findings

Two HIGH issues were found that need to be fixed before merge:

### [HIGH] Symlinks silently dropped — regression for editable installs and venvs

`DirEntry::file_type()` returns the type of the symlink itself (not its target), so `is_dir()` and `is_file()` both return false for symlinks. The previous code used `path.is_dir()`/`path.is_file()` which follow symlinks automatically. This would break scanning `site-packages` directories that contain symlinked `.so` extension modules (common in Linux distros and editable installs).

**Fix**: After `file_type()`, if `is_symlink()`, fall back to `entry.path().metadata()` (which follows symlinks).

### [HIGH] Namespace packages not emitted during scan

When a directory without `__init__.py` is found during scan, it is recursed but never reported as a `NamespacePackage` module. This is inconsistent with `find_in_path()` which does handle namespace packages.

**Fix**: Emit `NamespacePackage` entry for directories without `__init__.py`.

### Additional findings (MEDIUM/LOW)
- Recursive `thread::scope` with `PARALLEL_SUBDIR_THRESHOLD=10` could spawn unbounded threads for deep/wide trees → only parallelize at the top level
- Threshold of 10 may be too low and cause thread overhead for small scans

Fixing in the next commit.

PR #195, opened 2026-06-13
fix(run): add -c/--code flag for python -c compatibility (Issue #184)

### LINKED ISSUE #184: fix(run): `pybun run -c "code"` does not work like `python -c` (requires -c --)
## Summary
The `run` help text says `[TARGET] ... Use -c for inline code`, but the natural `python`-style invocation fails:

```
$ pybun run -c "print('hi')"
error: unexpected argument 'print('hi')' found
```

The only working form is to abuse `--` passthrough:
```
$ pybun run -c -- "print('hi')"   # works
```

## Root cause
`-c` is not a defined option; it is matched as the positional `TARGET` string `"-c"` (see `src/commands.rs:2466` `if target == "-c"`). clap then treats the following code string as an unexpected argument unless `--` is present.

## Impact
- Breaks muscle memory / `python -c` compatibility, and breaks AI agents that emit `pybun run -c "<code>"`.
- Help text is actively misleading.

## Proposed fix
Add a real option: `-c, --code <CODE>` on the `run` subcommand that takes the inline code as its value, and route it to `run_python_code`. Keep the legacy `-c --` form working for one release with a deprecation note, or remove it once `--code` lands. Update help text and snapshots.

## Test plan
- [ ] `pybun run -c "print('hi')"` prints `hi` and exits 0.
- [ ] `pybun run --code "import sys; sys.exit(7)"` exits 7.
- [ ] JSON mode reports child stdout/exit code as today.

takurot:
Closes #184

## Summary
- Add a proper `-c, --code <CODE>` option to `pybun run`, so `pybun run -c "print('hi')"` and `pybun run --code "..."` work like `python -c "..."`, matching AI-agent muscle memory.
- Previously `-c` was matched as the positional `TARGET`, so the code string was rejected as an unexpected extra argument unless callers abused `pybun run -c -- "code"` (passthrough escape hatch).
- The legacy `-c -- "code"` form is removed: with `-c` now defined as a value-taking option, clap cannot accept `--` as its value (it errors with "a value is required for '--code <CODE>' but none was supplied"). Updated all tests/docs that used the old form to the new `-c "code"` syntax.
- Updated `tests/snapshots/compat/help_run.txt` for the new `--code` option and help text.

## Test plan
- [x] `pybun run -c "print('hi')"` prints `hi` and exits 0 (`run_inline_code`)
- [x] `pybun run --code "import sys; sys.exit(7)"` exits 7 (`run_inline_code_long_flag` / `run_inline_code_propagates_nonzero_exit_code`)
- [x] JSON mode reports child stdout/exit code as before (`run_json_mode_inline_nonzero_reports_error_status`)
- [x] `cargo test` — all 332+ tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ pip install pybun-cli
 **Then run:**
 ```bash
 pybun add requests
-pybun run -c -- "import requests; print('Hello, PyBun!')"
+pybun run -c "import requests; print('Hello, PyBun!')"
 ```
 
 ---
@@ -91,7 +91,7 @@ Existing Python tools are built for **humans**. PyBun is designed for **both AI 
 $ pybun --format=json add pandas
 {"status": "ok", "detail": {"added": ["pandas==2.2.0"], ...}}
 
-$ pybun --format=json run -c -- "import pandas; print(pandas.__version__)"
+$ pybun --format=json run -c "import pandas; print(pandas.__version__)"
 {"status": "ok", "stdout": "2.2.0\n", ...}
 ```
 
@@ -166,7 +166,7 @@ pybun run script.py
 pybun run script.py -- arg1 arg2
 
 # Run inline code
-pybun run -c -- "import sys; print(sys.version)"
+pybun run -c "import sys; print(sys.version)"
 
 # Run with profile
 pybun run --profile=prod script.py
@@ -399,7 +399,7 @@ Requires `pip install pybun-cli`.
 All commands support the `--format=json` option (schema v1). Examples:
 
 ```bash
-pybun --format=json run -c -- "print('hello')"
+pybun --format=json run -c "print('hello')"
 ```
 
 ```json
--- a/docs/DEMO_GIF_GUIDE.md
+++ b/docs/DEMO_GIF_GUIDE.md
@@ -52,11 +52,11 @@ Type "pybun add requests httpx"
 Enter
 Sleep 2s
 
-Type "pybun run -c -- \"import requests; print('Hello from PyBun!')\""
+Type "pybun run -c \"import requests; print('Hello from PyBun!')\""
 Enter
 Sleep 1s
 
-Type "pybun --format=json run -c -- \"print('JSON output!')\""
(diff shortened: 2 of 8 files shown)


takurot:
## Code Review

Reviewed the diff (`src/cli.rs`, `src/commands.rs`, `src/entry.rs`, tests, docs) against `main`.

**Security (CRITICAL): none found**
- Inline code is passed as a single `arg("-c").arg(code)` to `ProcessCommand` — no shell involved, so no command-injection risk regardless of code content.
- No secrets/credentials touched.

**Code Quality (HIGH): none found**
- `run_python_code` signature change (`code: &str` param instead of reading `args.passthrough.first()`) is a clean, minimal refactor — removes a now-dead `ok_or_else` branch.
- No functions/files grew beyond size limits; no new nesting issues.

**Best Practices (MEDIUM):**
- Legacy `pybun run -c -- "code"` form is removed rather than kept as a deprecated alias, since clap can't accept `--` as an option value for `-c`. This is a deliberate, documented breaking change per the issue's accepted "or remove it" option — all affected tests (`tests/cli_run.rs`, `tests/verify_usability_fixes.rs`) and docs (`README.md`, `docs/DEMO_GIF_GUIDE.md`) were updated to the new `-c "code"` / `--code "code"` syntax. ✅ consistent.
- Help snapshot (`tests/snapshots/compat/help_run.txt`) regenerated via `PYBUN_UPDATE_SNAPSHOTS=1` — diff includes some unrelated reordering of `--progress`/`--allow-read`/`--no-progress` lines, which is just clap's recomputed arg ordering after adding the new `-c`/`--code` field, not a manual change.

**Tests:** New tests added (`run_inline_code_long_flag`) and existing ones (`run_inline_code`, `run_inline_code_propagates_nonzero_exit_code`, `run_json_mode_inline_nonzero_reports_error_status`, `test_env_source_display_labels`) updated to the new syntax. Full suite (`cargo test`), `cargo clippy -- -D warnings`, and `cargo fmt -- --check` all pass.

No CRITICAL or HIGH issues — looks good to merge pending CI.

PR #196, opened 2026-06-13
fix(resolver): PEP 508 marker evaluation incomplete (Issue #183)

### LINKED ISSUE #183: fix(resolver): PEP 508 marker evaluation incomplete (os_name, implementation_name, in/not in, quoted and/or)
## Summary
`evaluate_marker()` in `src/resolver.rs` only understands `platform_machine`, `sys_platform`, `platform_system`, and `python_version`. Every other PEP 508 marker variable and several operators fall through to an "inclusive true" default, which can pull in dependencies meant for other platforms/interpreters.

## Gaps (code: src/resolver.rs:935-1004)
- Unhandled variables → inclusive-true: `os_name`, `platform_python_implementation`, `implementation_name`, `platform_release`, `python_full_version`, `extra`.
- `in` / `not in` operators not implemented (TODO at :995) → inclusive-true.
- `and`/`or` splitting is quote-unaware (TODO at :944, :953) — a quoted value containing " or "/" and " is mis-split.

## Impact
- `; sys_platform == 'win32'`-style markers are handled, but `; os_name == 'nt'` is **not** → Windows-only deps can be selected on Linux/macOS, and vice versa.
- `extra == '...'` is handled in `src/pypi.rs::marker_allows` but NOT in `resolver.rs::evaluate_marker`, so the two marker evaluators disagree.

## Proposed fix
- Unify on a single PEP 508 marker evaluator (consider porting a small, tested implementation or `pep508_rs`).
- Implement the full env-marker variable set, `in`/`not in`, and a quote-aware tokenizer for `and`/`or`.
- Keep the inclusive-on-unknown fallback only for genuinely unknown variables, and emit a diagnostic when a marker can't be evaluated.

## Test plan
- [ ] Unit cases for each marker variable and operator (including `in`/`not in`).
- [ ] Quote-aware `and`/`or` split test (value containing " or ").
- [ ] Cross-check `resolver.rs` and `pypi.rs` agree on `extra ==`.

takurot:
Closes #183

## Summary
- Replaced the naive `evaluate_marker()` (string-split-based, quote-unaware `and`/`or`, only 4 supported variables) with a proper quote-aware tokenizer (`tokenize_marker`) and recursive-descent parser (`MarkerParser`) implementing the PEP 508 grammar: `marker_or := marker_and ('or' marker_and)*`, `marker_and := marker_expr ('and' marker_expr)*`, with parentheses for grouping.
- Added support for the full PEP 508 marker variable set: `os_name`, `sys_platform`, `platform_machine`, `platform_system`, `platform_release`, `platform_version`, `platform_python_implementation`, `implementation_name`, `python_version`, `python_full_version`, `implementation_version`.
- Implemented `in` / `not in` operators (substring containment per PEP 508 semantics).
- Quoted values containing `" or "` / `" and "` (e.g. `sys_platform == "linux or win32"`) are now tokenized correctly instead of being mis-split.
- `extra == "..."` now consistently evaluates to "extra not active" in both `resolver::evaluate_marker` and `pypi::marker_allows` (made `marker_allows` `pub(crate)` so the agreement can be tested directly). Unknown variables / unsupported operators / unparseable markers still fall back to inclusive (`true`) so packages are never silently dropped.
- Refactored Python version detection: `get_python_full_version()` now does the single subprocess call, and `get_python_version()` derives MAJOR.MINOR from it (no extra subprocess). Added `get_python_implementation()`, `get_platform_release()`, `get_platform_version()`, `get_os_name()`.

## Test plan
- [x] Unit tests for `os_name` (posix/nt)
- [x] Unit tests for `platform_python_implementation`/`implementation_name` (CPython/PyPy)
- [x] `in`/`not in` operator tests
- [x] Quote-aware `and`/`or` regression tests (value containing `" or "`/`" and "`)
- [x] `extra ==` agreement test between `resolver::evaluate_marker` and `pypi::marker_allows`
- [x] Unknown-variable / unparseable-marker inclusive fallback tests
- [x] Full existing test suite (`cargo test`) passes
- [x] `cargo fmt`


takurot:
## Code Review (manual review of `git diff main...HEAD`)

Reviewed `src/resolver.rs`, `src/pypi.rs`, `docs/PLAN.md`.

### Security
- No hardcoded secrets, no injection vectors. Subprocess calls (`python3 --version`, `uname -r`, `uname -v`) use fixed argument arrays (no shell interpolation), with fixed program names — no command injection risk.

### Code Quality
- Tokenizer (`tokenize_marker`) and recursive-descent parser (`MarkerParser`) cleanly implement the PEP 508 `marker_or`/`marker_and`/`marker_expr` grammar with quote-aware splitting and parentheses — addresses the prior `" or "`/`" and "` naive-split TODOs.
- All new functions are well-documented with rustdoc explaining fallback behavior.
- `OnceLock` caching avoids redundant subprocess spawns; `get_python_version()` now derives from `get_python_full_version()` instead of a separate subprocess call.
- `extra ==` handling is now consistent between `resolver::evaluate_marker` and `pypi::marker_allows` (both treat as "not active"), with the design rationale documented inline and in PLAN.md.
- Unknown variables / unparseable markers fall back to inclusive (`true`), preserving the "never silently drop packages" invariant from the original implementation.
- `unreachable!()` in `evaluate_comparison` is provably unreachable (guarded by the enclosing `">=" | "<=" | ">" | "<"` match) — acceptable.

### Minor / non-blocking observations (LOW, not addressed in this PR)
- `evaluate_marker` does not lowercase the marker string before comparison (unlike `pypi::marker_allows`, which lowercases everything). This is a pre-existing behavior, not a regression — e.g. `sys_platform == "Darwin"` (capitalized) wouldn't match `"darwin"` either before or after this change. Could be a follow-up if real-world packages are found using non-canonical casing.
- `get_platform_release()`/`get_platform_version()` return owned `String` (cloned from a cached `OnceLock<String>`) rather than `&'static str` — a minor allocation per call, acceptable given these are only evaluated per-marker during resolution.

### Tests
- 11 new tests cover: `os_name`, `platform_python_implementation`/`implementation_name`, `in`/`not in`, quote-aware `or`/`and` splitting (including literals containing `" or "`/`"and or"`), `extra ==`/`!=` parity with `pypi::marker_allows`, unknown-variable inclusivity, and unparseable-marker inclusivity.
- Full suite passes (341 lib tests + integration suites), `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt` clean.

**Verdict: No CRITICAL or HIGH issues found. Approved.**

PR #197, opened 2026-06-13
feat(watch): polling fallback for pybun watch (Issue #190)

### LINKED ISSUE #190: feat(watch): polling fallback so `pybun watch` runs without the native-watch feature
## Summary
In the default build (no `native-watch` feature), `pybun watch <file>` does not actually watch anything — it prints a "Would watch …" preview and exits:

```
$ pybun watch a.py
pybun watch: Would watch 1 paths for changes to run: a.py
...
Note: Native file watching requires building with --features native-watch.
```

Since release binaries are typically built without `native-watch`, `pybun watch` is effectively non-functional for most users. (This is the PR-A6 item in docs/PLAN.md; filing as a tracked issue.)

## Proposed
- Add a polling-based watcher fallback (mtime/size scan with the existing debounce + include/exclude pattern filtering) used when the native backend is unavailable.
- `pybun watch` should detect changes and re-run the command in **all** builds; native `notify` remains the fast path when compiled in.
- Keep `--shell-command` generation as an additional option.

## Acceptance
- [ ] In a default (no-feature) build, editing a watched file triggers a re-run (E2E, not just preview text).
- [ ] Debounce + include/exclude patterns honored by the polling path.
- [ ] Native-watch build still uses `notify`.

takurot:
## Summary
- Implements a poll-based file watcher (`scan_watch_paths`, `diff_snapshots`, `run_polling_watch_loop` in `src/hot_reload.rs`) used as the fallback for `pybun watch` when the `native-watch` feature (notify crate) is not compiled in — i.e. for release binaries.
- Replaces the previous `not(feature = "native-watch"))` path in `run_watch()` (which only printed a "Would watch... (requires native-watch)" preview) with a real blocking watch loop that detects file changes (mtime/size) and re-runs the target command.
- Reuses `should_watch_path` (extracted from `HotReloadWatcher::should_watch`) for include/exclude pattern filtering, and `debounce_ms` (min 100ms) as the polling interval.
- Adds `PYBUN_WATCH_MAX_ITERATIONS` test-only env var to bound the otherwise-infinite polling loop for E2E tests.
- Updates `docs/PLAN.md` to mark PR-A6 as `[DONE]`.

Closes #190

## Test plan
- [x] `cargo test --lib hot_reload::tests::polling_watch_tests` — 9 unit tests covering include/exclude filtering, snapshot scanning, diff detection (Created/Modified/Deleted), and loop behavior
- [x] `cargo test --test hot_reload` — E2E: polling fallback detects a file change and re-runs the configured command; invalid watch path produces a structured JSON error
- [x] `cargo test --features native-watch --test hot_reload` and `--lib hot_reload` — native-watch path still works (uses `notify` via `run_native_watch_loop`)
- [x] `cargo test` (full suite)
- [x] `cargo fmt -- --check`

## Acceptance criteria (Issue #190)
- [x] In a default (no-feature) build, editing a watched file triggers a re-run (E2E, not just preview text)
- [x] Debounce + include/exclude patterns honored by the polling path
- [x] Native-watch build still uses `notify`

### DIFF
--- a/docs/PLAN.md
+++ b/docs/PLAN.md
@@ -93,9 +93,10 @@
   - Current: `src/sandbox.rs::SITECUSTOMIZE_PY` の `_block_subprocesses()` に `os.posix_spawn`/`os.posix_spawnp`/`os.spawnv`/`os.spawnve`/`os.spawnvp`/`os.spawnvpe`/`os.spawnl`/`os.spawnle`/`os.spawnlp`/`os.spawnlpe`/`os.startfile`（Windows）を追加し、いずれも `_deny("process creation", "blocked_subprocesses")` を呼ぶよう `setattr` で上書き。
   - Tests: `tests/sandbox.rs` に2件追加（`sandbox_blocks_posix_spawn` — `os.posix_spawn`/`os.posix_spawnp` が `PermissionError` を発生させ `blocked_subprocesses:2` を記録し、エスケープ証跡ファイルが作成されないことを確認、`sandbox_blocks_spawn_family` — `os.spawnv`/`spawnve`/`spawnvp`/`spawnvpe`/`spawnl`/`spawnle`/`spawnlp`/`spawnlpe` の8系統がすべて `PermissionError` で `blocked_subprocesses:8` を記録することを確認）。全sandboxテスト(29件)・`cargo clippy --all-targets --all-features -- -D warnings`・`cargo fmt -- --check` がパス。
 
-- PR-A6: `pybun watch` の標準ビルド fallback 監視
+- [DONE] PR-A6: `pybun watch` の標準ビルド fallback 監視 (Issue #190)
   - Goal: `native-watch` 無効でも poll-based で監視再実行できるようにする。
-  - Tests: feature無効CIで実監視E2E（previewではなく変更検知→再実行）。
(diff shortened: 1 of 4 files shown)


takurot:
## Code review (automated)

**[HIGH] Shell command built via string interpolation (`format!("pybun run {}", target_script)`)**
`src/commands.rs:4793` (polling path, new in this PR) and `src/commands.rs:4747` (native-watch path, pre-existing) both build a shell command string from `target_script` (raw CLI arg) and execute it via `sh -c` / `cmd /C` in `src/hot_reload.rs`. A target path containing shell metacharacters or spaces could break or behave unexpectedly. This PR mirrors the existing native-watch pattern rather than introducing a new one — fixing it properly (shell-quoting or avoiding the shell) would affect both paths and is better tracked as a separate follow-up issue rather than scope-creeping this PR.

**[MEDIUM] `scan_watch_paths` follows symlinks without cycle detection**
A symlink loop inside a watched tree (not caught by default excludes) could cause unbounded recursion/memory growth per poll. Consider `symlink_metadata` + visited-path tracking in a follow-up.

**[MEDIUM] Full recursive rescan every poll interval**
Acceptable for a "fallback" feature per PLAN.md preview status; default excludes (`.git`, `.venv`, `node_modules`, etc.) mitigate the worst cases. Worth documenting as a known limitation for very large repos.

**[LOW] `PYBUN_WATCH_MAX_ITERATIONS` (test-only escape hatch)**
If accidentally set in a user's environment, `pybun watch` would silently stop after N iterations. Consider an info/warning log when honored, and document it as test-only (alongside other `PYBUN_*_TEST_*` vars).

**No CRITICAL issues.** Logic for snapshot diffing (Created/Modified/Deleted), include/exclude filtering, debounce-as-poll-interval clamping, and loop guards (empty/missing watch paths) all look correct, with good unit + E2E test coverage.

Plan: merge as-is (HIGH item is pre-existing/shared with native-watch path, not a regression; LOW/MEDIUM items are non-blocking polish for follow-up issues).

Issue #202, opened 2026-06-14
fix: stale pre-0.1.19 PyPI cache causes fatal E_RESOLVE_IO on add/install/lock; gc/doctor report wrong cache path


takurot:
## Summary

On a machine that previously used pybun <v0.1.19, the global PyPI metadata cache (`~/Library/Caches/pybun/pypi/*.bin` on macOS) contains bincode-serialized `CacheEntry` blobs in an old struct layout (without the `body: Vec<u8>` field added in v0.1.19). After upgrading to v0.1.20, `pybun add`/`install`/`lock`/`outdated`/`upgrade` fail with a fatal, misleading error instead of treating the stale cache entry as a cache miss and re-fetching.

A second, related issue: `pybun doctor`/`pybun gc` report and operate on a different cache directory than the one actually used for PyPI metadata, so there is no built-in way to detect or clean up the broken cache.

- Version: v0.1.20, commit `2049928be4513fe11c54d59451287bf6c8c4d011`
- Environment: macOS (Darwin 25.5.0, arm64), Python 3.11.15 (Homebrew)

### Reproduction

1. Have a `~/Library/Caches/pybun/pypi/requests.bin` (and transitive deps) written by pybun <v0.1.19 (old `CacheEntry` layout, no `body` field).
2. Run `pybun --format=json add requests`.

### Expected

`requests` and its dependencies resolve and install normally — a cache decode failure should be treated as a cache miss and trigger a fresh fetch from the index.

### Actual

```json
{"status":"error", ...}
```
with `E_RESOLVE_IO` and message:
```
io error: parse error: cache decode error: tag for enum is not valid, found 8
```
(or `found 7` for a different stale dependency cache file). The accompanying Japanese hint tells the user to check network connectivity or index file path/permissions — neither of which is the actual cause. The command fails completely; there is no automatic recovery.

### Root cause

- `src/pypi.rs:187` — `fetch_packages`: `let cached_entry = self.load_cache(name).await?;` propagates any cache error fatally.
- `src/pypi.rs:714-723` — `load_cache_from_paths`:
  ```rust
  let entry: CacheEntry = bincode::deserialize(&data)
      .map_err(|e| PyPiError::Parse(format!("cache decode error: {}", e)))?;
  ```
[...]

PR #204, opened 2026-06-14
fix(pypi): treat stale pre-0.1.19 bincode cache as miss; surface PyPI cache in doctor/gc (Issue #202)

takurot:
## Summary

Fixes issue #202, a HIGH severity regression introduced in v0.1.19.

**Part A (fatal regression fix):** `CacheEntry` gained a `body: Vec<u8>` field in v0.1.19. Bincode `.bin` cache entries written by older pybun versions now fail to deserialize, and this decode error was propagated as a fatal `E_RESOLVE_IO`, breaking `pybun add`/`install`/`lock`. Such decode failures are now treated as a cache miss (re-fetch from index) with a `warning` diagnostic surfaced to the caller.

**Part B (doctor/gc cache coverage):** `pybun doctor`/`pybun gc` only reported `~/.cache/pybun` (`Cache::new()`), but the actual PyPI metadata cache lives at `dirs::cache_dir()/pybun/pypi` (e.g. `~/Library/Caches/pybun/pypi` on macOS) and was not covered by either command.

## Changes

- `src/pypi.rs`:
  - `load_cache_from_paths` now returns `(Option<CacheEntry>, Option<String>)` — a bincode decode error on an existing `.bin` is `(None, Some(notice))` instead of a `PyPiError`.
  - `PyPiClient` gains `stale_cache_notices: Arc<Mutex<Vec<String>>>` and `take_stale_cache_notices()` (also exposed via `PyPiIndex`).
  - New `pypi_cache_dir()`, `pypi_cache_stats()`, `gc_stale_pypi_cache()` helpers.
- `src/commands/mod.rs`:
  - `install`, `lock_dependencies`, PEP 723 native run, `run_outdated`, and `upgrade` drain stale-cache notices into `collector.warning(...)`.
  - `run_doctor` adds a `pypi_cache` check (entry_count/total_bytes/stale_count), with an info diagnostic + `pybun gc` hint when stale entries exist.
  - `run_gc` merges PyPI cache GC results into `freed_bytes`/`files_removed`/`would_remove` and adds a `pypi_cache` JSON detail section.
- `tests/gc.rs`: new integration tests for dry-run/non-dry-run `gc` and `doctor` against a stale PyPI cache directory.
- `src/pypi.rs` unit tests: new `stale_bincode_cache_is_treated_as_cache_miss` regression test (pre-`body`-field bincode layout).
- `docs/PLAN.md`: added PR-A9 entry documenting the fix.

## Test plan

- [x] `cargo test --lib pypi::` — 5/5 pass, including new regression test
- [x] `cargo test --test gc` — 10/10 pass, including 3 new tests
- [x] `cargo test --test e2e_general` — 9/9 pass
- [x] `cargo test` — full suite passes


takurot:
## Code review (automated)

Ran the code-reviewer subagent on this PR. Summary of findings and resolutions:

| Severity | Finding | Resolution |
|----------|---------|------------|
| HIGH | Notice wording ("discarded incompatible PyPI cache entry...") asserts a specific root cause (old layout) that the code can't verify — a bincode decode failure could also be a truncated write from a crash. | Fixed in f3098d4: reworded to "discarded unreadable PyPI cache entry..." and softened the doc comment to acknowledge both causes. |
| MEDIUM | `pypi_cache_stats` and `gc_stale_pypi_cache` duplicated the "is this `.bin` stale" check. | Fixed in f3098d4: factored shared `is_stale_pypi_cache_entry()` helper used by both. |
| MEDIUM | `run_upgrade`'s resolve call used `?` before draining stale-cache notices, so notices were lost on a resolve error (the exact scenario where they'd be most useful). | Fixed in f3098d4: reordered to drain notices before propagating the error, matching the other 4 call sites. |
| MEDIUM | `gc_stale_pypi_cache` swallows `fs::remove_file` errors silently. | Accepted as-is — matches existing GC code style in this file (best-effort cleanup); not introduced by this PR. |
| LOW | `Arc<Mutex<Vec<String>>>` notice pattern is a latent footgun if a cloned `PyPiClient` is reused across multiple resolve passes before draining. | Accepted — not an issue for current single-drain-per-command usage; scope of a larger refactor is out of scope for this fix. |
| NIT | `PYBUN_PYPI_CACHE_DIR` trust / doc comment precedence vs `PYBUN_HOME` | Accepted — mirrors existing `cache.gc()` trust assumptions; no behavior change needed. |

All tests (`cargo test`, 352 lib tests + integration suites), `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo fmt -- --check` pass after the fixes.

Issue #186, opened 2026-06-11
refactor(commands): split src/commands.rs (6841 lines) into per-command modules


takurot:
## Summary
`src/commands.rs` is 6,841 lines — by far the largest file in the tree and ~8.5x the project's own 800-line guideline. It is the dispatcher plus the implementation of nearly every command, which makes review, navigation, and merge-conflict avoidance hard.

## Current state
- `src/commands.rs`: 6841 lines (next largest is `mcp.rs` at 1772).
- Most subcommand logic (`run`, `install`, `lock`, `test`, `build`, `x`, `gc`, `doctor`, sandbox wiring, etc.) lives inline in one file.

## Proposed approach (incremental, low-risk)
- Keep `commands.rs` as a thin `execute()` dispatcher.
- Extract per-domain modules: `commands/run.rs`, `commands/install.rs`, `commands/lock.rs`, `commands/test.rs`, `commands/build.rs`, `commands/x.rs`, `commands/maintenance.rs` (gc/doctor), etc.
- Move the inline `#[cfg(test)] mod tests` blocks alongside their code.
- No behavior change; pure code movement verified by the existing suite.

## Test plan
- [ ] `cargo test` unchanged green before/after each extraction.
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` clean.
- [ ] No public API/JSON changes (compat snapshots unchanged).

PR #221, opened 2026-06-21
fix(runtime): exclude pandas/matplotlib from lazy-import default denylist (Issue #136)

takurot:
Closes #136

## Summary
- Benchmarks (B6.1/B6.2/B6.3) showed `pandas` and `matplotlib` are slightly **slower** with lazy-import enabled (1.05x–1.08x), because their own CPython import cost is already low relative to the fixed overhead of the lazy-import `sys.meta_path` hook.
- Added `pandas` and `matplotlib` to `src/lazy_import.rs::default_denylist()` so they import eagerly by default. `numpy` is unaffected and keeps its measured 1.87x speedup under lazy-import.
- Denylist takes priority over allowlist (existing behavior), so users can still force lazy-loading via `pybun lazy-import --allow pandas` if desired.

## Test plan
- [x] Unit tests: `test_default_denylist_excludes_modules_with_no_lazy_benefit`, `test_should_lazy_import_denies_pandas_and_matplotlib` (new, Red→Green)
- [x] Updated pre-existing unit tests that used `pandas` as an "allowed" example (`test_should_lazy_import_allowed_module`, `test_allowlist_mode`) to use `scipy`/`numpy`/`requests` instead
- [x] Integration test: `tests/lazy_import.rs::test_lazy_import_check_pandas_denied` (new); updated `test_lazy_import_check_json` to use `scipy`
- [x] `docs/PLAN.md` updated with PR-A10 entry

## Follow-up (not in this PR, per Issue #136 suggestions)
- Per-module timing breakdown in `pybun lazy-import --show-config`
- Smarter eager-access-pattern detection for the denylist heuristic

### DIFF
--- a/docs/PLAN.md
+++ b/docs/PLAN.md
@@ -254,6 +254,10 @@ Milestones follow SPECS.md Phase roadmap. PR numbers are suggested grouping; par
   - Depends on: PR2.1.  
   - Current: `src/lazy_import.rs` implements lazy import configuration with allowlist/denylist, default denylist for core modules (sys, os, importlib, etc.). Python code generation for `sys.meta_path` injection with LazyModule proxy, LazyFinder, LazyLoader classes. `pybun lazy-import` command with `--generate`, `--check`, `--show-config`, `--allow`, `--deny`, `--log-imports`, `--no-fallback` options. Config file support (TOML).
   - Tests: 17 unit tests (config, allowlist/denylist logic, stats, serialization); 12 E2E tests (help, config, check, generate, file output).
+- [DONE] PR-A10: lazy-import default denylist excludes pandas/matplotlib (Issue #136)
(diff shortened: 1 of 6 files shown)


takurot:
**Automated code review (code-reviewer subagent)** found and I've fixed:

- **[HIGH]** `docs/PLAN.md` reused the ID `PR-A10`, already in use by the Issue #186 entry. Renamed the new entry to `PR-A12` (next free ID after `PR-A11`).
- **[MEDIUM]** `scripts/benchmark/config.toml` / `scripts/benchmark/scenarios/lazy_import.py` still listed `pandas`/`matplotlib` in the default `heavy_modules` for the B6.1 lazy-import benchmark. Since they're now eagerly imported by default, the benchmark would silently stop measuring lazy-import behavior for them. Updated default to `["numpy"]` with an explanatory comment.
- **[LOW]** `docs/SPECS.md` lazy-import description used Pandas as the example heavy library; updated to NumPy + a note that Pandas/Matplotlib are denylisted by default (Issue #136).

All fixes pushed in a follow-up commit. Full `cargo test`, `cargo clippy -- -D warnings`, `cargo fmt --check`, and `python3 -m unittest discover -s scripts/benchmark/tests` pass locally.

PR #277, opened 2026-07-01
fix(sandbox): warn when credential-shaped --allow-env is rejected (Issue #259)

### LINKED ISSUE #259: security: --allow-env sandbox flag is a no-op — allow-listed vars never reach subprocess
## Summary
`pybun run --sandbox --allow-env=<VAR>` does not actually pass the allow-listed environment variable into the sandboxed subprocess. This silently defeats the escape hatch added by #153 for passing non-secret config (e.g. `PYBUN_PROFILE`) into sandboxed AI-generated code.

## Repro (verified on `main` @ 09c35b7, v0.1.21)
```bash
SECRET_API_KEY=topsecret pybun --format=json run --sandbox --allow-env=SECRET_API_KEY -c "import os; print('SECRET_API_KEY' in os.environ)"
```

**Expected:** stdout `True`, and `detail.sandbox.allow_env` in the JSON envelope should list `["SECRET_API_KEY"]`.

**Actual:** stdout `False` — identical to a run without `--allow-env` at all. `detail.sandbox.allow_env` is `[]` even though `--allow-env=SECRET_API_KEY` was explicitly passed on the command line.

Full JSON output observed:
```json
{"detail":{"sandbox":{"allow_env":[],"allow_network":false, ...},"stdout":"False\n", ...}}
```

## Impact
Since env-var filtering is default-deny (correctly blocking secrets), this bug means there is currently **no working way** to pass any environment variable into `--sandbox` mode. Any workflow that legitimately needs one non-secret config value (e.g. `PYBUN_PROFILE`, a feature flag, a locale setting) inside sandboxed AI-generated code is currently unsupported despite the documented `--allow-env` flag.

## Suggested area
`src/sandbox.rs` — allow-list parsing/threading into the spawned subprocess environment.

takurot:
Closes #259

## Summary

- Investigated the reported repro (`--allow-env=SECRET_API_KEY` reaches the subprocess as `False`) and confirmed the *env filtering itself* is working as designed: `is_credential_env_name` intentionally blocks any allow-listed name that looks like a credential (`AWS_*`, `*_SECRET`, `*_TOKEN`, `*_KEY`) as defense-in-depth, even if explicitly requested. Verified non-credential-shaped names (e.g. `MY_FLAG`) already pass through correctly both via `-c` and script execution paths.
- The real defect: this rejection was **silent**. `detail.sandbox.allow_env` just showed `[]` with zero diagnostic explaining why the caller's explicit `--allow-env` request was dropped — indistinguishable from a total no-op bug.
- Added `SandboxGuard::rejected_env` (`src/sandbox.rs`) populated from `config.allow_env` filtered by `is_credential_env_name`, and a new `emit_rejected_allow_env_diagnostics()` helper in `src/commands/mod.rs` that emits a `W_SANDBOX_ALLOW_ENV_REJECTED` warning diagnostic per rejected name, wired into both sandboxed run paths (script file and inline `-c`).
- No change to the actual security behavior — credential-shaped names are still never passed through, allow-listed or not.

## Test plan
- [x] `cargo test --lib sandbox::` — 3 new unit tests covering allow/reject/mixed cases
- [x] `cargo test --test sandbox` — new integration test asserting `W_SANDBOX_ALLOW_ENV_REJECTED` + var name appear in JSON diagnostics
- [x] `cargo test --test mcp`
- [x] `cargo fmt -- --check`
- [x] Manual repro of issue's exact command now shows the warning diagnostic instead of silently returning `allow_env: []`


takurot:
## Code review (rust-reviewer subagent)

**Verdict: Approve**

- `cargo fmt --check`, `cargo check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --lib sandbox::`, `cargo test --test sandbox` all pass.
- **Correctness**: `is_credential_env_name` filtering is unchanged; `rejected_env` is derived via the exact inverse predicate, so the split is exhaustive and existing security behavior is unaffected (purely additive). The diagnostic uses `Diagnostic::warning(...).with_code("W_SANDBOX_ALLOW_ENV_REJECTED")`, matching the sibling `emit_unsupported_resource_limit_diagnostics` pattern, so it correctly surfaces as a warning rather than an error.
- **Call-site consistency**: Both `run_script` and `run_python_code` call `emit_rejected_allow_env_diagnostics` immediately after `emit_unsupported_resource_limit_diagnostics`, in the same relative position — consistent.
- **Tests**: unit tests cover pass-through-only, reject-only, and mixed cases; integration test checks for the code + var name in JSON.

**Minor, non-blocking**: `rejected_env` is only exposed via the `diagnostics` array, not added to `detail.sandbox`. Since the original bug report referenced `detail.sandbox.allow_env`, agents parsing structured `detail` rather than `diagnostics` won't see rejected names directly there. This is an intentional design choice (diagnostics is the structured channel for actionable warnings/errors in this codebase's JSON schema), so no change requested, flagging for visibility only.

PR #278, opened 2026-07-01
fix(cli): support multiple packages in pybun add/remove (Issue #264)

### LINKED ISSUE #264: fix: pybun add fails when passed multiple package arguments
## Summary
`pybun add` only accepts a single package name positional argument. Passing multiple packages (a very common workflow, e.g. `pip install requests rich`) fails with a clap parsing error instead of adding all of them.

## Repro (verified on `main` @ 09c35b7, v0.1.21)
```bash
mkdir demo && cd demo && pybun init -y
pybun add requests rich
```

**Actual:**
```
error: unexpected argument 'rich' found
```

**Expected:** `pybun add requests rich` should resolve and install both packages in one invocation, matching `pip install`/`uv add`/`npm install` conventions.

## Impact
Basic ergonomics gap that will surprise nearly every new user coming from pip/uv/npm; forces N separate `pybun add` invocations (N resolver passes) instead of one batched resolve.

## Suggested area
CLI arg definition for the `add` subcommand — likely `src/cli.rs` or wherever `add`'s clap `Args` struct is defined (single `String` positional instead of `Vec<String>`), plus the corresponding handler in `src/commands/mod.rs`/`src/commands/install.rs` to loop over multiple package specs in one resolve pass.

takurot:
Closes #264

## Summary
- `pybun add`/`pybun remove` previously accepted only a single package positional argument (`Option<String>`), so `pybun add requests rich` failed with a clap "unexpected argument" error.
- `PackageArgs.package: Option<String>` is now `PackageArgs.packages: Vec<String>` (required, at least one value), matching pip/uv/npm ergonomics.
- `add_package`/`remove_package` now loop over all provided package specs, updating `pyproject.toml` once and running a single resolve/install pass for `add`.
- JSON output keeps the existing `package`/`version`/`removed` fields (mirroring the first package) for backward compatibility, and adds a new `packages` array with per-package detail.
- `remove` reports partial success when some packages are removed and others are not found (e.g. `removed requests; nonexistent was not found in dependencies`).

## Test plan
- [x] `cargo test --test cli_add_remove` — includes new `add_accepts_multiple_packages`, `remove_accepts_multiple_packages`, `remove_multiple_reports_partial_not_found` tests
- [x] `cargo test --test json_schema`
- [x] `cargo test --test compat_snapshots` (updated `help_add.txt`/`help_remove.txt` snapshots for the new `<PACKAGE>...` usage)
- [x] `cargo fmt`
- [x] Manually reproduced the issue's repro steps (`pybun init -y && pybun add requests rich`) and confirmed both packages are resolved/installed in one invocation; verified `pybun remove requests rich` too.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -183,9 +183,10 @@ pub struct LockArgs {
 
 #[derive(Args, Debug)]
 pub struct PackageArgs {
-    /// Package name (optionally with version).
-    #[arg(value_name = "PACKAGE")]
-    pub package: Option<String>,
+    /// Package name(s) (optionally with version). Multiple packages may be
+    /// given in a single invocation.
+    #[arg(value_name = "PACKAGE", required = true)]
+    pub packages: Vec<String>,
     /// Use offline mode when cache is sufficient.
     #[arg(long)]
     pub offline: bool,
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -128,12 +128,16 @@ pub async fn execute(cli: Cli) -> Result<()> {
             match result {
                 Ok(AddOutcome {
                     summary,
-                    package,
-                    version,
+                    packages,
                     added_deps,
                 }) => {
                     // Chain install to ensure the environment is up-to-date
-                    collector.info(format!("Installing dependencies including {}...", package));
+                    let names = packages
+                        .iter()
+                        .map(|p| p.name.as_str())
+                        .collect::<Vec<_>>()
+                        .join(", ");
+                    collector.info(format!("Installing dependencies including {}...", names));
 
                     let install_args = crate::cli::InstallArgs {
                         offline: args.offline,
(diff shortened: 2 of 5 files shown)


takurot:
## Code review (subagent)

Ran an independent code-reviewer pass on the diff. Summary:

- **CRITICAL/HIGH:** none found.
- **MEDIUM (fixed in 99304dc):** duplicate package names in one `pybun add` invocation (e.g. `pybun add numpy numpy==2.0`) produced duplicate/stale entries in the JSON `packages` array, even though `Project::add_dependency` only persists the last spec per name to `pyproject.toml`. Fixed by deduping `packages` to keep the last occurrence per name, matching what's actually saved. Added a regression test (`add_deduplicates_repeated_package_name`).
- **LOW (no change, pre-existing behavior):** `pybun remove` exits 0 even if some requested packages aren't found — callers must inspect `packages[].removed` rather than relying on exit code. This matches prior single-package behavior and is out of scope for this fix.
- Atomicity: `add_package` validates all package specs and only calls `project.save()` once after the loop, so a mid-loop parse failure never leaves `pyproject.toml` partially written.

PR #287, opened 2026-07-03
fix(bootstrap): fail closed on placeholder checksum and HTTP asset URLs (Issue #283)

### LINKED ISSUE #283: security: pip-shim bootstrap silently skips checksum verification on placeholder sha256 (fail-open)
### Summary
`pybun/bootstrap.py` (the pip/pipx shim that downloads and executes the release binary) **silently skips checksum verification** when the manifest's `sha256` field is the literal `placeholder`. The Rust side treats the same situation as a hard error, so the Python shim is the weak link in the release verification chain.

### Details

**1. Placeholder checksum silently disables verification** — `pybun/bootstrap.py:238-245`

```python
def _verify_checksum(path: str, expected: Optional[str]) -> None:
    if not expected:
        raise BootstrapError("manifest missing sha256 for asset")
    expected_clean = expected
    if expected_clean.startswith("sha256:"):
        expected_clean = expected_clean[len("sha256:") :]
    if expected_clean in {"sha256:placeholder", "placeholder"}:
        return  # <-- silently accepts the binary unverified
```

A manifest that carries `"sha256": "placeholder"` (or `sha256:placeholder`) results in the downloaded archive being extracted and **executed with no integrity check and no warning**. The Rust self-update path explicitly rejects this (`src/self_update.rs:148-149`):

```rust
if normalized == "placeholder" || normalized == "sha256:placeholder" {
    return Err(err("placeholder checksum is not allowed for self update"));
}
```

The shim should fail closed the same way (or at minimum require an explicit opt-out), not fail open.

**2. Dead branch**: after the `sha256:` prefix strip on line 242-243, the set member `"sha256:placeholder"` on line 244 can never match — evidence the placeholder handling was not reasoned through.

**3. Plain-HTTP asset URLs are accepted** — `bootstrap.py:225` allows `http://` asset URLs. Combined with (1), a MITM'd manifest/asset pair runs arbitrary code. Assets should be HTTPS-only (or local-file for dev).

**4. Tar extraction doesn't use `filter="data"`** — `_safe_extract_tar` checks `member.name` for path traversal but not `linkname`; a tar containing a symlink pointing outside the destination followed by a member writing through it can escape. Python ≥3.12 supports `tar.extractall(dest, filter="data")`, which handles this class of attacks.

### Expected behavior
- Placeholder/empty checksums are rejected with a clear error, matching `self_update.rs` semantics.
- Verification can only be skipped via the explicit, documented `PYBUN_PYPI_NO_VERIFY=1` escape hatch (which should also print a loud warning).
- Asset downloads require HTTPS.
- Tar extraction uses `filter="data"` where available.

### Impact
Anyone installing via `pip install pybun-cli` / pipx relies on this shim as the trust root for the binary they run. Fail-open verification undermines the ed25519/minisign signing work already done on the Rust side (PR-A1).

### Environment
- Repo state: `main` @ 07f1a7a, v0.1.21

takurot:
Closes #283

## Summary
- `pybun/bootstrap.py` is the pip/pipx shim that downloads and executes the PyBun release binary — it is the trust root for `pip install pybun-cli` users. It previously **silently skipped checksum verification** when the manifest's `sha256` field was the literal `"placeholder"`, while the Rust `self_update.rs` path already rejects this case with a hard error. This patch makes the shim fail closed to match.
- `_download_url` now only accepts `https://` (and `file://` for local/dev use); plain `http://` asset URLs are rejected with a clear error instead of being silently fetched.
- `_safe_extract_tar` now also validates symlink/hardlink `linkname` targets (not just member paths) against the destination directory, and opts into `tarfile.extractall(..., filter="data")` on Python 3.12+ as defense in depth.
- `ensure_binary()` prints a loud stderr warning when the existing `PYBUN_PYPI_NO_VERIFY=1` escape hatch is used, so skipping verification is no longer silent.

## Test plan
- [x] `pytest pybun/tests/test_bootstrap.py` — 16 passed (7 pre-existing + 9 new: placeholder/empty/mismatch/match checksum cases, http rejection / file allowance, tar path-traversal / symlink-escape / valid-extraction cases)
- [x] `cargo test --test pypi_shim` — existing Rust integration test (uses `file://` + real sha256) still passes unmodified
- [x] `cargo test` (full workspace) — all passing
- [x] `ruff check pybun/bootstrap.py pybun/tests/test_bootstrap.py` — clean
- [x] `docs/PLAN.md` updated with PR-A19 entry


takurot:
## Code review (code-reviewer subagent)

| Severity | Count |
|---|---|
| CRITICAL | 0 |
| HIGH | 1 |
| MEDIUM | 1 |
| LOW | 2 |

**[HIGH]** `load_manifest()` (bootstrap.py:135) still allows plaintext `http://` for the *manifest* fetch. The asset-download HTTPS restriction added in this PR can be bypassed by attacking the manifest fetch instead (attacker-controlled manifest supplies both a malicious HTTPS asset URL *and* a matching sha256/signature). Same trust-root threat model as the rest of this PR — will fix here.

**[MEDIUM]** Symlink escape check resolves `linkname` relative to `dest_abs` instead of the symlink's own containing directory, which can false-positive-reject legitimate nested symlinks (e.g. `bin/lib -> ../shared/lib`). Not an exploitable bypass (only over-rejects), but a correctness bug — will fix.

**[LOW]** `import sys` inline inside `ensure_binary()` instead of top-level import block — will fix.

**[LOW]** `_verify_checksum` raises raw `AttributeError` instead of `BootstrapError` if manifest's `sha256` field is a non-string JSON value — will fix with an `isinstance` check.

Test gaps identified: hardlink escape case, legitimate-nested-symlink case, `PYBUN_PYPI_NO_VERIFY=1` stderr warning assertion, malformed non-string `sha256` case. Adding all four.

Addressing all findings in a follow-up commit on this branch.

takurot:
Addressed all review findings in 1752cc0: manifest fetch now rejects plaintext HTTP (closing the bypass of the asset HTTPS restriction), symlink target resolution fixed to use the link's own directory (fixes false-positive rejection of legitimate nested symlinks; verified with a new test), non-string sha256 now raises a clean BootstrapError, and the inline `import sys` was moved to the top-level imports. Added 5 more tests (26 total, all passing) covering: manifest http rejection, NO_VERIFY stderr warning, hardlink escape rejection, legitimate nested symlink acceptance, non-string checksum rejection.

PR #288, opened 2026-07-05
fix(install): default install target no longer falls back to system Python (Issue #286)

### LINKED ISSUE #286: fix(install): default install target falls back to system Python site-packages (PR-A7)
### Summary
When no virtual environment exists, `pybun install` resolves the environment down to **system Python and installs wheels directly into its `site-packages`**, printing only a warning. CLAUDE.md and docs/PLAN.md track this as PR-A7 ("install の安全な既定ターゲット"), but there is no GitHub issue for it — filing one so it's visible in the tracker alongside the other audit follow-ups.

### Details

- `src/env.rs` environment discovery order: `PYBUN_ENV` → active venv → project `.pybun/venv`/`.venv`/`venv` → **`find_system_python()` fallback** (`src/env.rs:139`).
- `src/commands/mod.rs:1580-1626`: when the resolved env is `EnvSource::System` and a `pyproject.toml` exists, it prints a warning + hint — then **proceeds to compute system `purelib` and unzip wheels into it**:
  ```rust
  let warning = "warning: PyBun is installing into system Python but a pyproject.toml exists.";
  ...
  // Determine site-packages path  → sysconfig.get_paths()['purelib']  → install_wheel(...)
  ```
- With no `pyproject.toml` there is not even a warning.

### Why this matters
- Polluting system/Homebrew/pyenv global site-packages is exactly the failure mode PyBun's own docs warn about ("Never install directly to system Python without explicit user intent" — CLAUDE.md, Installation Safety).
- On PEP 668 (`EXTERNALLY-MANAGED`) distros, pip refuses this by default; PyBun bypasses that guard entirely because it unzips wheels itself.
- Agent-first tooling makes this worse: an agent running `pybun install` in a fresh checkout will silently mutate the host Python.

### Expected behavior (PR-A7)
- Default: if no project venv exists, create `.pybun/venv` (or `.venv`) and install there.
- Installing into system Python requires an explicit flag (e.g. `--system`), mirroring uv's behavior.
- Respect `EXTERNALLY-MANAGED` when the system target is explicitly requested.

### Related
- PR-A7 in docs/PLAN.md (line ~141), CLAUDE.md "Installation Safety" constraint

### Environment
- Repo state: `main` @ 07f1a7a, v0.1.21

takurot:
Closes #286

## Summary
- `pybun install` previously fell back to system Python's site-packages whenever no venv/`PYBUN_ENV` was found, printing only a warning (and no warning at all if `pyproject.toml` was absent). This directly violated the documented Installation Safety constraint ("Never install directly to system Python without explicit user intent") and bypassed PEP 668 `EXTERNALLY-MANAGED` protections that `pip` respects by default.
- Default behavior now auto-creates and installs into a project-local `.pybun/venv` when no venv is found (`src/env.rs::create_project_venv`).
- Added an explicit `--system` flag (`InstallArgs`) to opt into installing into the resolved system Python. When used, PyBun checks for the PEP 668 `EXTERNALLY-MANAGED` marker via `src/env.rs::externally_managed_marker` and refuses with a new `E_INSTALL_EXTERNALLY_MANAGED` diagnostic if present.
- `pybun add` (which chains into `install()` internally) keeps the safe default (`system: false`).

## Test plan
- [x] `src/env.rs` unit tests: `create_project_venv_creates_pybun_venv_directory`, `create_project_venv_reuses_existing_venv`, `externally_managed_marker_detects_pep668_file`, `externally_managed_marker_absent_when_no_marker_file`
- [x] `tests/pypi_integration.rs` new E2E tests exercising the *real* download+install path via a local httpmock PyPI server (not the `--index` fixture path, which never populates real download URLs and thus never exercises this code):
  - `install_creates_project_local_venv_instead_of_system_python_by_default` — confirms `.pybun/venv` is auto-created and the wheel lands there, not in the PATH-resolved "system" interpreter
  - `install_system_flag_installs_into_resolved_system_python` — confirms `--system` installs into the resolved system interpreter with an explicit warning, and does not create `.pybun/venv`
- [x] `cargo test --lib` (421 passed)
- [x] `cargo test --test cli_install --test pypi_integration --test compat_snapshots --test cli_add_remove --test json_schema` (all passed; updated `tests/snapshots/compat/help_install.txt` for the new `--system` flag)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` (clean)
- [x] `cargo fmt -- --check` (clean)
- [x] `docs/PLAN.md` PR-A7 marked `[DONE]`; `CLAUDE.md` Installation Safety / Known Issues sections updated

Note: `cli_run`/PEP 723 tests fail locally due to a pre-existing broken dev `.venv` in this checkout (linked against a since-removed Homebrew Python 3.14 dylib) — unrelated to this change and reproducible on `main` before this branch.


takurot:
### Code review (post-merge)

Ran an independent review of this diff after merge. Found 1 real gap and two smaller issues:

1. **[HIGH] `.python-version` fallback bypasses this PR's safety guard entirely.**

   `install()` only guards `EnvSource::System` ([src/commands/mod.rs#L1580](https://github.com/VOID-TECHNOLOGY-INC/PyBun/blob/e55b906fb528ac2934678eb802f4adeb82bb32f8/src/commands/mod.rs#L1580)). But `find_python_env()`'s `.python-version` branch ([src/env.rs#L120-L144](https://github.com/VOID-TECHNOLOGY-INC/PyBun/blob/e55b906fb528ac2934678eb802f4adeb82bb32f8/src/env.rs#L120-L144)) calls `find_python_for_version()`, which falls back to a bare PATH lookup when no pyenv-managed version matches ([src/env.rs#L246-L271](https://github.com/VOID-TECHNOLOGY-INC/PyBun/blob/e55b906fb528ac2934678eb802f4adeb82bb32f8/src/env.rs#L246-L271)) and tags the result `EnvSource::PythonVersionFile`, not `System`. A project with a `.python-version` file and no matching pyenv install resolves to the same real system interpreter this PR was meant to guard — but skips the auto-created `.pybun/venv`, the `--system` check, and the externally-managed guard entirely. This reproduces Issue #286's exact failure mode through a different code path. Filed as a follow-up: (see next comment / issue link).

2. **[MEDIUM] `externally_managed_marker()` fails open on subprocess errors** ([src/env.rs#L383-L395](https://github.com/VOID-TECHNOLOGY-INC/PyBun/blob/e55b906fb528ac2934678eb802f4adeb82bb32f8/src/env.rs#L383-L395)) — a failed probe (spawn error, non-UTF8 output) is treated as "not externally managed" rather than "unknown." Lower severity since `--system` already requires explicit opt-in.

3. **[MEDIUM] `create_project_venv` uses `.status()` instead of `.output()`** ([src/env.rs#L348-L351](https://github.com/VOID-TECHNOLOGY-INC/PyBun/blob/e55b906fb528ac2934678eb802f4adeb82bb32f8/src/env.rs#L348-L351)) — lets the child `python -m venv` process's stdout/stderr flow directly through, unlike every other subprocess call touched in this PR; risks corrupting the single-JSON-document contract `--format=json` depends on if the venv bootstrap ever prints to stdout.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #290, opened 2026-07-05
fix(install): .python-version bare-PATH fallback bypassed safe-install-target guard (Issue #289)

### LINKED ISSUE #289: bug(install): .python-version fallback bypasses safe-install-target guard from Issue #286
### Summary

PR #288 (Issue #286) fixed `pybun install` silently falling back to system Python by guarding on `EnvSource::System` in `install()` (`src/commands/mod.rs:1580`) — auto-creating `.pybun/venv` by default and requiring an explicit `--system` flag (with a PEP 668 `EXTERNALLY-MANAGED` check) to install into system Python.

However, `find_python_env()`'s `.python-version` resolution branch can resolve to the exact same real, unmanaged system interpreter but tag it `EnvSource::PythonVersionFile` instead of `EnvSource::System` — bypassing the new guard entirely.

### Details

- `src/env.rs:120-144`: when a `.python-version` file is present, `find_python_for_version(&version)` is called.
- `src/env.rs:246-271` (`find_python_for_version`): tries pyenv first (`find_pyenv_python`), then falls back to a bare PATH lookup — `which_executable("python3.11")` / `which_executable("python3")` — if no pyenv-managed install matches.
- If that PATH lookup succeeds (e.g. `.python-version` contains `3.11` and the host has a system `python3.11` but no pyenv), the result is wrapped as `EnvSource::PythonVersionFile(version_file)`, not `EnvSource::System`.
- `install()`'s guard (`src/commands/mod.rs:1580`) only matches `matches!(env.source, EnvSource::System)`. `PythonVersionFile` results skip the auto-created `.pybun/venv`, the `--system` requirement, and the `E_INSTALL_EXTERNALLY_MANAGED` check — installing directly into the resolved system interpreter with no warning at all.

### Reproduction (expected)

1. In a project with no `.pybun/venv`/`.venv`/`PYBUN_ENV`, add a `.python-version` file containing a version that isn't pyenv-installed but matches a system `python3.X` on `PATH` (e.g. `3.11` on a host with system `python3.11`).
2. Run `pybun install`.
3. Expected (per Issue #286): refuse, or auto-create `.pybun/venv`, or require `--system`.
4. Actual: installs directly into the system interpreter resolved via `.python-version`, with no warning.

### Expected behavior

Extend the safe-install-target logic in `install()` to also cover `EnvSource::PythonVersionFile` results whose resolved interpreter isn't inside a project-local venv (e.g. key the check off "is `python_path` inside a directory PyBun created/controls" rather than solely the `EnvSource` enum tag), or otherwise ensure `find_python_env` distinguishes a pyenv-isolated install from a bare PATH-resolved fallback.

### Related

- #286 (original issue), PR #288 (fix, incomplete for this path)
- Found via post-merge code review of PR #288.

takurot:
Closes #289

## Summary
- PR #288 (Issue #286) fixed `pybun install` silently falling back to system Python by guarding `install()` on `EnvSource::System`. Independent post-merge review of #288 found that `find_python_env()`'s `.python-version` branch could resolve to the exact same kind of unmanaged, PATH-found interpreter but tag it `EnvSource::PythonVersionFile` instead — silently bypassing the new guard (no `.pybun/venv` auto-creation, no `--system` requirement, no `EXTERNALLY-MANAGED` check).
- `src/env.rs::find_python_for_version()` now returns `(PathBuf, bool)` where the `bool` indicates whether resolution went through a pyenv-managed, per-version install (`true`) or fell back to a bare `which python3.X`/`python3` PATH lookup (`false`).
- `find_python_env()` now tags the bare-PATH-lookup case as `EnvSource::System` (matching reality — it's just the system interpreter, discovered via a version hint), so it automatically receives the same safe-install-target handling as the plain system-Python fallback. The pyenv-isolated case keeps its existing `EnvSource::PythonVersionFile` tag.

## Test plan
- [x] New unit test `python_version_file_bare_path_lookup_is_tagged_as_system` in `src/env.rs`: writes a `.python-version` file matching the running system Python's version (with no pyenv install for it) and asserts `find_python_env()` returns `EnvSource::System`. Skips gracefully if pyenv is detected on the host (targets the non-pyenv fallback branch specifically) or if no system Python is available.
- [x] `cargo test --lib` (422 passed)
- [x] `cargo test --test cli_install --test pypi_integration --test compat_snapshots --test cli_add_remove --test json_schema` (all passed — confirms no regression to the Issue #286 fix or install-command behavior)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` (clean)
- [x] `cargo fmt -- --check` (clean)
- [x] `docs/PLAN.md` PR-A7 entry updated with this follow-up


takurot:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #292, opened 2026-07-05
fix(install): select wheels for the target venv's Python, not PATH (Issue #291)

### LINKED ISSUE #291: bug(resolver): off-by-one wheel python_tag selection - resolves cp311 into Python 3.12+ venv (regression of #161 on v0.1.21)
## Summary

`pybun install` on **v0.1.21** (latest stable) resolves `cp311` wheels into a Python 3.12 venv, producing `ModuleNotFoundError: No module named 'numpy.core._multiarray_umath'` at runtime. Follows the off-by-one pattern of #161 (target 3.11 -> cp310), now manifesting as target 3.12 -> cp311.

## Environment

- **PyBun**: 0.1.21 (channel: stable, `pybun-cli` 0.1.21, `pybun self update` reports up-to-date)
- **OS**: macOS 15.x (Darwin 25.5.0), Apple Silicon (aarch64-apple-darwin)
- **Target Python**: 3.12.7 (managed by `pybun python install`)
- **Wheel**: numpy 1.26.4 cp311 tag installed; a cp312 arm64 wheel EXISTS on PyPI but was NOT selected
- **Host Python**: 3.14.6 (homebrew, used by shim only)

## Reproduction

```
# 1. Create project
mkdir repro && cd repro
cat > pyproject.toml << 'PYEOF'
[project]
name = "repro"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = ["numpy<2.0.0"]
PYEOF

# 2. Set up a Python 3.12 venv
PY312=$(pybun python which 3.12.7)
$PY312 -m venv .venv

# 3. Install via pybun
export PYBUN_ENV=.venv
export PYBUN_PYTHON=$PY312
pybun install

# 4. Verify the installed wheel tag
find .venv/lib -name WHEEL -path '*numpy*' -exec grep '^Tag:' {} \;
# Output: Tag: cp311-cp311-macosx_11_0_arm64
# Expected: cp312-cp312-macosx_11_0_arm64

# 5. Import fails
.venv/bin/python -c "import numpy"
# ImportError: No module named 'numpy.core._multiarray_umath'
```

## JSON resolution evidence (`pybun install --format=json`)

Selected artifacts have the wrong python_tag:

- contourpy-1.3.3-**cp311**-cp311-macosx_11_0_arm64.whl
- faiss_cpu-1.14.3-**cp310-abi3**-macosx_14_0_arm64.whl (stable ABI, not affected)
- numpy-1.26.4-**cp311**-cp311-macosx_11_0_arm64.whl

## Controls tested - none override the wrong tag

| Control | Result |
|---|---|
| `requires-python = ">=3.12"` in pyproject.toml | Still cp311 |
| `PYBUN_PYTHON=<3.12.7 interp>` | Still cp311 |
| `PYBUN_PYPI_PYTHON_VERSION=3.12` | Still cp311 |
| `.python-version = "3.12.7"` | Still cp311 |
| venv created with Python 3.12.7 | Still cp311 |

## Off-by-one pattern

- **#161 (v0.1.18)**: target Python 3.11 -> resolved `cp310` wheels
- **This issue (v0.1.21)**: target Python 3.12 -> resolves `cp311` wheels

The fix in v0.1.21 (commits 87980bb5, 1233e7b4, d4a1a0ef) appears to have shifted the off-by-one upward by one Python minor version rather than eliminating it.

## Expected behavior

PyBun should detect the target venv Python version (3.12) and select `cp312` wheels from PyPI when they exist.

## Severity

Critical for any project using `pybun install` with Python 3.12+ and per-version packages (numpy, contourpy, pyarrow, etc.). Workaround: use `pip install` with a standard venv.

## Related

- #161 - original ABI resolution bug (closed)
- #144 - platform tag matching bug (closed)
- aLayaSiki#55 - downstream Issue blocked by this PyBun regression

takurot:
Closes #291

## Summary

- `pybun install` selected wheel `python_tag`s via `active_python_cp_tag()`, which shells out to whatever `python3`/`python` resolves on `PATH`. This ignored `PYBUN_ENV`, `PYBUN_PYTHON`, `.python-version`, and the project venv entirely when picking wheels — so wheel selection could pick a CPython ABI that doesn't match the interpreter packages are actually installed into (e.g. `cp311` wheels resolved into a Python 3.12 venv), producing `ImportError: No module named 'numpy.core._multiarray_umath'` at runtime.
- Moved install-target resolution (`find_python_env` + the system-Python safe-install-target guard) earlier in `install()`, before wheel selection, and derived the active CPython tag directly from that interpreter's actual `--version` output.
- Both wheel-selection passes (lockfile write and download list) now use `select_artifact_for_platform_with_cp(..., &active_cp_tag)` instead of the PATH-based `select_artifact_for_platform`.
- `PYBUN_FORCE_CP_TAG` still overrides detection, preserving existing test determinism.

## Test plan

- [x] New regression test `install_selects_wheel_for_target_venv_python_not_path_python` in `tests/cli_install.rs`: builds a fake venv reporting `Python 3.12.5` (independent of the host's real Python/PATH) and asserts the resolved lockfile picks the `cp312` wheel over a decoy `cp311` wheel.
- [x] `cargo test --test cli_install` — 24 passed
- [x] `cargo test` (full suite) — same pass/fail counts as `main` except for pre-existing, unrelated `cli_run` PEP 723 failures caused by a broken local Python 3.14 framework install on this dev machine (verified identical failures on `main` before this change)

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1426,41 +1426,17 @@ async fn install(
         event.progress = Some(40);
     });
 
-    // Resolve the actual install target (PYBUN_ENV / PYBUN_PYTHON / project venv /
-    // system Python) *before* selecting wheels, so artifact selection matches the
-    // Python interpreter packages will actually be installed into. Selecting wheels
-    // against whatever `python3`/`python` happens to resolve on PATH (the previous
-    // behavior) can silently pick wheels for the wrong CPython ABI (Issue #291).
+    // Detect the CPython tag of the actual install target (PYBUN_ENV / PYBUN_PYTHON /
+    // project venv / system Python) *before* selecting wheels, so artifact selection
+    // matches the Python interpreter packages will actually be installed into.
+    // Selecting wheels against whatever `python3`/`python` happens to resolve on PATH
+    // (the previous behavior) can silently pick wheels for the wrong CPython ABI
+    // (Issue #291). This is read-only detection only — creating a project-local venv
+    // (and the associated system-Python safe-install-target guard) is deferred to the
+    // later "Install wheels" step below, so a resolve-only or failed install doesn't
+    // have the side effect of mutating the filesystem.
     let working_dir = std::env::current_dir()?;
-    let mut env = crate::env::find_python_env(&working_dir)?;
-
-    if matches!(env.source, crate::env::EnvSource::System) {
-        if args.system {
-            if let Some(marker) = crate::env::externally_managed_marker(&env.python_path) {
-                let message = format!(
-                    "refusing to install into externally-managed system Python (marker: {})",
-                    marker.display()
-                );
-                collector.error_with_code(
-                    "E_INSTALL_EXTERNALLY_MANAGED",
-                    message.clone(),
-                    "This interpreter is marked externally-managed (PEP 668). Create a virtual environment (e.g. `python3 -m venv .venv`) and re-run, or install with a non-managed interpreter.",
-                );
-                return Err(eyre!(message));
-            }
-
-            let warning =
-                "warning: PyBun is installing into system Python (--system was specified).";
-            eprintln!("{}", warning);
-            collector.warning(warning.to_string());
-        } else {
-            collector.info(
-                "No virtual environment found; creating project-local environment at .pybun/venv"
-                    .to_string(),
-            );
-            env = crate::env::create_project_venv(&working_dir)?;
-        }
-    }
+    let target_env_probe = crate::env::find_python_env(&working_dir)?;
 
     // PYBUN_FORCE_CP_TAG lets tests (and users) pin the CPython tag deterministically,
     // bypassing interpreter detection entirely.
@@ -1468,7 +1444,7 @@ async fn install(
         .ok()
         .filter(|v| !v.trim().is_empty())
         .or_else(|| {
-            get_python_version(&env.python_path)
+            get_python_version(&target_env_probe.python_path)
                 .ok()
                 .and_then(|v| python_version_to_cp_tag(&v))
         })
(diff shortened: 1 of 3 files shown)


takurot:
## Code review (self-review via code-reviewer subagent)

Reviewed the diff against `origin/main`. Summary:

- **Core fix is correct**: computing `active_cp_tag` from the resolved target interpreter (`find_python_env`) instead of PATH-based `python3 --version`, threaded through `select_artifact_for_platform_with_cp` at both wheel-selection call sites. New regression test (`install_selects_wheel_for_target_venv_python_not_path_python`) verified to fail on pre-fix code and pass on the fix.
- **[HIGH, fixed]**: the first version of this fix moved `find_python_env`/`create_project_venv` earlier to compute the cp tag, which made venv creation (and the system-Python safe-install-target guard) run unconditionally on every install — including resolve-only and failure paths. This was a real regression (unlocked concurrent `.pybun/venv` creation across parallel tests, filesystem mutation on failed installs). Fixed in a follow-up commit: cp-tag detection now uses a read-only `find_python_env()` probe, while the mutating venv-creation/guard logic stays at its original later position, gated on `!download_items.is_empty()`.
- **[MEDIUM, follow-up]**: `lock_dependencies`, `run_script`, and `run_upgrade` still use the PATH-based `select_artifact_for_platform` — the same class of bug as #291 could still affect `pybun lock` / `pybun run` (PEP 723) / `pybun upgrade`. Filing as a follow-up rather than expanding this PR's scope.
- No security issues found (no new subprocess-injection surface; `env.python_path` comes from the same trust boundary as before).

Verified `cargo test --test cli_install` (24 passed), `cargo clippy --all-targets --all-features -- -D warnings` (clean), `cargo fmt -- --check` (clean).

PR #329, opened 2026-07-09
fix(pep723): self-heal last_used cache decode failures (Issue #306)

### LINKED ISSUE #306: fix: update_last_used_at in pep723_cache.rs has same decode-fatal pattern as #299, left unfixed
## Summary

Issue #299 fixed `Pep723Cache::read_cache_entry` (`src/pep723_cache.rs`) so that an undecodable `deps.json` is treated as a cache miss (`Ok(None)`) instead of propagating a fatal error. The same file has a sibling function, `Pep723Cache::update_last_used_at` (~line 393), that reads and decodes the *same* `deps.json` file with the identical unguarded pattern:

```rust
let content = fs::read_to_string(&info_path)?;
let mut info: CachedEnvInfo = serde_json::from_str(&content)?;
```

This was not touched by the #299 fix, so the same bug class (decode failure treated as fatal instead of self-healing) still exists in this function.

## Why it hasn't caused a visible crash (yet)

Both current call sites swallow the `Result` outright:

- `get_cached_env` (~line 240): `let _ = self.update_last_used(hash);`
- `get_cached_env_fast` (~line 282): `let _ = self.update_last_used_at(cache_root);`

So today a corrupt `deps.json` just silently fails to bump `last_used` rather than crashing `pybun run`. But this is fragile:

- It's inconsistent with the just-established self-heal convention in the same file/struct — a future refactor that stops swallowing the error (e.g. to surface a real I/O failure) would silently reintroduce a fatal crash on cache corruption, exactly like #299.
- `last_used` drives LRU eviction ordering (see `list_cached_envs` sorting at ~line 533, used for cache GC). A cache entry that can never get its `last_used` refreshed due to a swallowed decode error stays invisible to `list_cached_envs` (which already skips undecodable entries via `if let Ok(info) = ...`), so the corrupt venv directory is neither refreshed nor evicted — it lingers on disk until something else touches it via `read_cache_entry`'s new self-heal path.

## Suggested fix

Apply the same self-heal pattern used in the #299 fix for `read_cache_entry`: on `serde_json::from_str` failure, log an informational notice (e.g. "discarded unreadable PEP 723 cache entry at {path} ({err}); skipping last_used update") and return `Ok(())` (a no-op) instead of propagating a `Pep723CacheError`. This keeps the function's behavior honest and consistent with `read_cache_entry` in the same struct.

## Suggested area

`src/pep723_cache.rs::update_last_used_at` (~line 393), and its wrapper `update_last_used` (~line 387).

## Related

- #299 (fixed the sibling `read_cache_entry` decode path — this issue covers the one function in the same file it didn't touch)
- #262 (same pattern, PyPI legacy cache)
- #301 (same self-heal-inconsistency pattern, `Lockfile::from_bytes` callers)

Found during a follow-up sweep for bugs similar to #299/#300, checked against `origin/main`.

takurot:
Closes #306

## Summary
- Treat corrupt PEP 723 deps.json in update_last_used_at as a non-fatal no-op, matching read_cache_entry self-heal behavior.
- Add a regression test that forces the decode path with an old mtime and verifies corrupt JSON does not propagate a fatal error.

## Test plan
- cargo fmt -- --check
- cargo clippy --all-targets --all-features -- -D warnings
- cargo test update_last_used_at_self_heals_on_corrupt_json
- cargo test
- cargo audit (passes with existing allowed RUSTSEC-2026-0190 warning)

## Review
- reviewer subagent: APPROVE, no blocking correctness or security findings.

### DIFF
--- a/src/pep723_cache.rs
+++ b/src/pep723_cache.rs
@@ -428,7 +428,17 @@ impl Pep723Cache {
             }
 
             let content = fs::read_to_string(&info_path)?;
-            let mut info: CachedEnvInfo = serde_json::from_str(&content)?;
+            let mut info: CachedEnvInfo = match serde_json::from_str(&content) {
+                Ok(info) => info,
+                Err(e) => {
+                    eprintln!(
+                        "info: discarded unreadable PEP 723 cache entry at {} ({}); skipping last_used update",
+                        info_path.display(),
+                        e
+                    );
+                    return Ok(());
+                }
+            };
 
             let now = SystemTime::now()
                 .duration_since(SystemTime::UNIX_EPOCH)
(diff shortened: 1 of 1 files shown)


takurot:
Code review: APPROVE\n\nReviewer subagent found no blocking correctness or security issues. Residual note: coverage is focused on update_last_used_at directly rather than an end-to-end fast-cache-hit flow, which is acceptable for this latent helper-level issue because callers currently swallow the update result. Local gates passed: fmt check, clippy -D warnings, targeted test, full cargo test, cargo audit with existing allowed RUSTSEC-2026-0190 warning.

Issue #239, opened 2026-06-27
perf: dependency resolver is 30x slower than uv — architectural overhaul needed


takurot:
## Problem

Benchmark results from #236 / #237:

| Scenario | pybun | uv | Ratio |
|----------|-------|----|-------|
| C5_resolution (small project, 10 pkgs) | 748ms p50 | 24ms p50 | **31x slower** |
| C5_resolution (min observed) | 551ms | 19ms | **29x slower** |
| C5_resolution (max observed) | 1602ms | 466ms | **3x–30x range** |

uv resolves a 10-package project in **24ms**. pybun takes **748ms** for the same project.

For a 50-package project (not yet measured), the gap will likely widen further given quadratic scaling of greedy resolution algorithms.

## Root Cause

From `CLAUDE.md` and `src/resolver.rs`:

> The resolver uses a **greedy highest-version selection strategy**. It:
> 1. Parses version specifiers
> 2. Fetches candidate versions from index
> 3. **Filters by version constraints**
> 4. Selects highest compatible version
> 5. Resolves transitive dependencies **recursively**

This is a naive recursive resolver with no parallelism and O(n²) behavior on dependency graphs.

**uv's approach:**
- Uses [PubGrub](https://github.com/pubgrub-rs/pubgrub) — a SAT-solver-inspired algorithm
- Fetches package metadata **in parallel** using async I/O (multiple packages simultaneously)
- Caches metadata aggressively to avoid re-fetching on subsequent lock operations
- Uses compact binary representation for version sets

## What 30x slower means in practice

For an AI agent that needs to add a package:

```bash
# Agent wants to add httpx to an existing project
pybun add httpx   # ← ~750ms (current)
uv add httpx      # ← ~25ms
```

This makes `pybun add` feel sluggish in interactive agent loops where dependency changes are frequent.

### Phase 1: Parallel metadata fetching (High impact, Medium effort)
The biggest win is fetching package metadata for multiple packages concurrently. Currently each package's metadata is fetched serially.

```rust
// Current (serial)
for dep in dependencies {
    let metadata = fetch_metadata(&dep).await?;  // blocks
}

// Target (parallel)
let metadata: Vec<_> = futures::stream::iter(dependencies)
    .map(|dep| fetch_metadata(&dep))
    .buffer_unordered(16)  // 16 concurrent fetches
    .collect()
    .await;
```

### Phase 2: Aggressive metadata caching
[...]

PR #331, opened 2026-07-09
perf(resolver): parallelize metadata fetching (Issue #239 Phase 1)

takurot:
## Summary

Part of #239 — this is **Phase 1 only** of the multi-phase resolver overhaul proposed in that issue. Phases 2 (disk caching), 3 (PubGrub solver integration), and 4 (lazy resolution) are **not** addressed here and remain follow-up work.

The frontier-based resolve loop in `src/resolver.rs::resolve()` already fetched sibling packages' *version lists* (`PackageIndex::all`) concurrently via `futures::future::try_join_all`. However, the second, more expensive network-bound step — fetching full metadata/dependencies for each *selected* package via `PackageIndex::get` — ran strictly serially inside a synchronous `for` loop, one package at a time, even for independent sibling packages discovered at the same recursion level. This serial `get()` step is the primary contributor to the 30x slowdown vs `uv` described in #239.

### Changes
- `src/resolver.rs`: split frontier processing into three passes:
  1. **Synchronous version selection** — unchanged ordering/semantics for constraint accumulation and conflict detection (no I/O).
  2. **Concurrent `index.get()` fetch** — all packages selected in this batch are fetched in parallel via `futures::stream::iter(...).buffer_unordered(MAX_CONCURRENT_METADATA_FETCHES)` (constant = 16, per the issue's example).
  3. **Synchronous commit** — inserts results into `resolved`, enqueues dependencies for the next frontier, preserving the pre-existing (slightly inconsistent) marker-filtering behavior between the "new package" and "conflict reconciliation" code paths exactly as before.
  - The existing `index.all()` fetch was also switched from unbounded `try_join_all` to the same bounded `buffer_unordered` pattern for consistency and to cap fan-out.
- `tests/resolver_basic.rs`: added `fetches_sibling_dependency_metadata_concurrently`, which uses an artificially delayed test `PackageIndex` to prove (a) sibling `get()` fetches now overlap in wall-clock time (tracks max concurrent in-flight fetches, asserts `> 1`), and (b) resolution results remain correct and deterministic despite non-deterministic completion order of the delayed fetches.

Resolution output (selected versions, conflict errors, dependency propagation) is unchanged — only the I/O for independent sibling fetches now overlaps in time instead of running one-at-a-time.

## Benchmark

Measured on the same machine (Apple M1, macOS 25.5.0), `scripts/benchmark/scenarios/uv_comparison.py` C5 scenario (offline fixture index, small 10-package project), before/after via `git stash`:

| | pybun install p50 |
|---|---|
| Before (main) | ~148–154 ms |
| After (this PR) | ~137–141 ms |

This offline/in-memory fixture scenario shows a modest ~7–10% improvement, since the fixture index has near-zero per-call latency to amortize — the serial-vs-parallel savings scale with actual per-fetch latency (e.g. real PyPI network round-trips), which this fixture doesn't simulate. The included regression test uses an artificial per-fetch delay (50ms × 8 siblings) to directly demonstrate the concurrency win under network-like conditions: resolution completes in well under half of the serial lower bound.

## Test plan
- [x] `cargo test --test resolver_basic` — 19/19 passed (18 existing + 1 new concurrency regression test)
- [x] `cargo test --test pypi_integration` — 9/9 passed
- [x] `cargo test` (full suite) — 993 passed, 6 ignored (same as baseline)
- [x] `cargo fmt` — clean, no diff
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — no issues
- [x] `cargo build --release` — succeeds
- [x] Benchmark comparison (`scripts/benchmark/bench.py -s uv_comparison`) captured before/after on this machine

### DIFF
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -742,17 +742,24 @@ pub async fn resolve(
 
         // 2. Fetch version-list metadata for sibling packages in parallel, bounded to
         // MAX_CONCURRENT_METADATA_FETCHES concurrent requests (Issue #239 Phase 1).
+        //
+        // Fail-fast note: `buffer_unordered` still runs up to
+        // MAX_CONCURRENT_METADATA_FETCHES fetches concurrently, but we drive the
+        // stream with a manual `while let` loop instead of collecting the whole
+        // batch first. This returns to the caller as soon as the first error is
+        // observed (whichever fetch happens to complete first — not necessarily
+        // submission order) instead of waiting for every in-flight fetch to
+        // finish. Futures already queued but not yet polled are dropped (and
+        // therefore cancelled) when we return early.
         if !names_to_fetch.is_empty() {
-            let results: Vec<Result<(String, Vec<ResolvedPackage>), ResolveError>> =
+            let mut stream =
                 futures::stream::iter(names_to_fetch.into_iter().map(|name| async move {
                     let pkgs = index.all(&name).await?;
                     Ok::<(String, Vec<ResolvedPackage>), ResolveError>((name, pkgs))
                 }))
-                .buffer_unordered(MAX_CONCURRENT_METADATA_FETCHES)
-                .collect()
-                .await;
+                .buffer_unordered(MAX_CONCURRENT_METADATA_FETCHES);
 
-            for result in results {
+            while let Some(result) = stream.next().await {
                 let (name, pkgs) = result?;
                 version_cache.insert(name, pkgs);
             }
@@ -762,10 +769,22 @@ pub async fn resolve(
         // this batch, which package version should be selected. This must stay
         // sequential because constraint accumulation (`constraints`) and
         // conflict detection depend on processing order — but it performs no
-        // I/O, so it's fast. `batch_resolved` mirrors what would land in
-        // `resolved` so later requirements in the same batch that target an
-        // already-selected package go through the same reconciliation logic
-        // the original serial implementation used.
+        // I/O, so it's fast.
+        //
+        // `fetch_events` records one entry per *selection event* in strict
+        // processing order (not deduped by package name). This matters for
+        // diamond dependencies: if two requirements in the same batch target the
+        // same newly-seen package with different constraints (e.g. `c<3.0.0` and
+        // `c<1.6.0`), the original serial resolver would select+fetch the first
+        // candidate (`c==2.0.0`), queue *its* dependencies, and only then
+        // reconcile to the second, narrower candidate (`c==1.5.0`) and queue
+        // *its* dependencies too — leaving both sets of dependencies in
+        // `next_batch` even though only the reconciled version ends up in
+        // `resolved`. Deduping by name (as an earlier version of this function
+        // did) silently drops the first candidate's dependencies, changing the
+        // resolved package set. Keeping a ordered Vec of events and replaying
+        // them in order after the concurrent fetch reproduces that exact
+        // behavior while still fetching metadata concurrently.
         enum FetchKind {
             /// Newly selected package — dependencies are filtered by marker on push.
             New,
@@ -773,8 +792,18 @@ pub async fn resolve(
             /// dependencies are pushed unfiltered, matching prior behavior.
             Reconcile,
         }
+        struct FetchEvent {
+            name: String,
+            candidate: ResolvedPackage,
+            requested_by: Option<String>,
+            kind: FetchKind,
+        }
+        // Tracks the latest selection per package name within this batch, purely
+        // for constraint-satisfaction lookups by subsequent requirements in the
+        // same batch (mirrors what `resolved` would contain in the serial
+        // implementation at each point in the loop).
         let mut batch_resolved: BTreeMap<String, ResolvedPackage> = BTreeMap::new();
-        let mut fetch_queue: BTreeMap<String, (Option<String>, FetchKind)> = BTreeMap::new();
+        let mut fetch_events: Vec<FetchEvent> = Vec::new();
 
         for (req, requested_by) in &current_batch {
             constraints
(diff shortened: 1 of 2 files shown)


takurot:
## Code review findings (rust-reviewer subagent, verified)

**Recommendation: changes requested before merge.**

### HIGH — Unstated resolution-output behavior change for diamond dependencies (`src/resolver.rs` ~769-893)

Splitting frontier processing into sync-select / concurrent-fetch / sync-commit passes silently changes the *set* of resolved packages when two same-batch requirements target the same newly-seen package with different constraints (diamond dependency shape).

Repro: `appA` requires `c<3.0.0`, `appB` requires `c<1.6.0`, both discovered in the same batch. `c@2.0.0` (first-picked) depends on `leftover-dep`; the final reconciled version is `c@1.5.0`, which does not depend on `leftover-dep`.

- On `main`: `resolution.packages` includes `leftover-dep` (deps of the superseded first-picked version were already queued before reconciliation).
- On this PR: `fetch_queue`/`batch_resolved` are `BTreeMap`s keyed by package name, so the second (reconcile) entry silently overwrites the first before its dependencies are ever fetched — `leftover-dep` disappears from the output.

Both branches select the same final version of `c`, and neither errors, but the resolved package set differs. This may be a legitimate bug fix (old behavior arguably pollutes the graph with deps of a version never actually selected) but it's an untested, unannounced semantic change bundled into a PR framed as pure concurrency. Please either:
1. Explicitly document this as an intentional behavior change and add a dedicated regression test for it, or
2. Adjust the refactor to preserve the exact prior dependency-push semantics.

### MEDIUM — Loss of fail-fast error propagation, non-deterministic error selection (`src/resolver.rs` ~743-759, 842-867)

`index.all()` and `index.get()` batches switched from `try_join_all` (short-circuits on first error) to `stream::iter(...).buffer_unordered(N).collect()`, which drives the whole batch to completion regardless of errors, then surfaces whichever error happens to be first in the *completion-ordered* `Vec`. This means: (a) a single slow fetch in a batch now delays error reporting even if another fetch already failed, and (b) which error is returned when multiple fetches fail is non-deterministic across runs. Worth a conscious, commented decision (e.g. use `TryStreamExt::try_collect` for fail-fast-while-still-bounded, if that tradeoff is acceptable) rather than an implicit side effect of the combinator swap.

### LOW — `expect()` on an internal invariant (`src/resolver.rs:851`, `874`)

Both `expect("every fetch_queue entry has a selected candidate")` calls are currently sound (verified), but panic instead of returning a typed error if the `fetch_queue`/`batch_resolved` key-parity invariant is ever broken by a future edit. Consider a typed internal error or `debug_assert!` + graceful handling for consistency with the project's no-panic-on-recoverable-state convention.

### No issues found
- Determinism of resolution *output*: confirmed correct — `BTreeMap` iteration in the sync-commit pass is independent of `buffer_unordered` completion order.
- `MAX_CONCURRENT_METADATA_FETCHES = 16`: reasonable, well-placed, no conflicting concurrency knobs elsewhere.
- New test `fetches_sibling_dependency_metadata_concurrently`: well-constructed (uses an atomic high-water-mark counter, not just timing, to prove overlap), CI-safe margins. Does not cover the HIGH-severity diamond-dependency gap above.

takurot:
Addressed the review feedback in f2a151d.

**HIGH — diamond dependency resolved-set change**: Chose the preferred option — preserve `main`'s exact serial resolution-output semantics rather than shipping the accidental change. Root cause: `fetch_queue`/`batch_resolved` were `BTreeMap<String, _>` keyed by package name, so when two same-batch requirements targeted the same newly-seen package with different constraints (e.g. `c<3.0.0` then `c<1.6.0`), the second (reconciling) selection silently overwrote the first before its dependencies were ever fetched.

Fix: replaced the name-keyed maps with an ordered `Vec<FetchEvent>` that records one entry per *selection event* (not deduped by name) in the exact order the old serial `for (req, requested_by) in current_batch` loop would have processed them. Metadata for all events is still fetched concurrently (bounded by `MAX_CONCURRENT_METADATA_FETCHES`), but events are replayed in original order afterward — so a package selected twice in one batch (New, then Reconcile) gets `index.get` called twice, both candidates' dependencies get queued into `next_batch`, and `resolved`/`parents` end up with the last (reconciled) version, exactly like `main`.

Added `diamond_dependency_reconciliation_preserves_first_candidate_deps` in `tests/resolver_basic.rs`, which constructs the exact repro from the review (appA requires `c<3.0.0`, appB requires `c<1.6.0`, `c==2.0.0` depends on `leftover-dep`, `c==1.5.0` doesn't). I verified the expected baseline by running the equivalent test against `main` directly (in a throwaway worktree) before writing the assertions — confirmed `main` resolves `c` to `1.5.0` while still including `leftover-dep` in `resolution.packages` (4 packages total: appA, appB, c, leftover-dep). The new test asserts this exact set/count and passes on this branch.

**MEDIUM — fail-fast error propagation**: Replaced `stream.collect::<Vec<Result<_,_>>>()` with a manual `while let Some(result) = stream.next().await { ... }` loop that returns on the first `Err` observed, for both the `index.all()` batch and the `index.get()` batch. Concurrency is still bounded via `buffer_unordered(MAX_CONCURRENT_METADATA_FETCHES)`. This doesn't cancel futures already polled into the buffer, but does stop consuming further items and returns to the caller immediately once an error surfaces (already-in-flight-but-unpolled futures are dropped when the stream is dropped on early return) — documented inline with a comment explaining the tradeoff.

**LOW — expect() on internal invariant**: Resolved as a side effect of the `fetch_events` rewrite — there's no longer a `fetch_queue`/`batch_resolved` key-parity invariant to assert. `fetched[idx].take().unwrap_or(event.candidate)` now falls back to the pre-fetch candidate instead of panicking if `index.get` ever returns `None`.

**Verification**: `cargo test` — 994 passed, 6 ignored, 0 failed (50 suites, 146.43s). `cargo clippy --all-targets --all-features -- -D warnings` — no issues. `cargo fmt -- --check` — clean.

Commit: f2a151d

PR #352, opened 2026-07-12
fix(downloader): fail fast on 4xx and preserve retry cause (Issue #343)

### LINKED ISSUE #343: fix(downloader): 4xx responses are retried with backoff, and MaxRetriesExceeded discards the underlying error cause
## Problem

Two related defects in `Downloader::download_file_with_signature` (`src/downloader.rs:144-192`):

### 1. Non-retryable HTTP errors are retried with backoff

`download_attempt` maps *every* failure — including `error_for_status()` on 4xx — into `DownloadError::Network`, and the retry loop retries all errors uniformly:

- A `404` (package/version genuinely missing, private index without auth → `401`/`403`) burns 3 attempts with 1s + 2s backoff ≈ **3+ seconds of guaranteed-futile waiting** before the user sees an error.
- Under `download_parallel` with many artifacts against a misconfigured index, this multiplies across every file.

Retry should be reserved for transient failures (connect/timeout errors, 5xx, 429); 4xx (except 429) should fail fast.

### 2. The final error discards the root cause

```rust
// src/downloader.rs:181-184
return Err(DownloadError::MaxRetriesExceeded {
    url: url.to_string(),
});
```

The last underlying error `e` is only ever `eprintln!`-ed on intermediate retries. The terminal error — the one that becomes the CLI/JSON diagnostic — says only "max retries exceeded for {url}". For a tool whose selling point is structured, agent-readable diagnostics, the actual cause (DNS failure? 403? checksum mismatch on re-download? disk full?) is exactly what the agent needs to pick a fix.

## Expected

1. Classify errors: fail fast on 4xx (except 408/429); retry connect/timeout/5xx/429 with the existing backoff.
2. Carry the last error in the terminal variant, e.g. `MaxRetriesExceeded { url, attempts, source: Box<DownloadError> }`, and surface it in the rendered message and JSON diagnostics.
3. (Optional) Respect `Retry-After` on 429/503.

## Affected code

- `src/downloader.rs:144-213` (retry loop, `download_attempt`, error enum)
- Diagnostics rendering wherever `DownloadError` is converted (install/upgrade paths, MCP `pybun_install`)

takurot:
Closes #343

## Summary
`Downloader::download_file_with_signature` had two related defects:

1. **Non-retryable HTTP errors were retried with backoff.** `download_attempt` mapped every failure — including a 4xx from `error_for_status()` — into a single retryable bucket. A `404` (missing package/version) or `401`/`403` (private index without auth) burned all 3 attempts with 1s + 2s of guaranteed-futile backoff before the caller saw an error. Under `download_parallel` against a misconfigured index this multiplied across every artifact.
2. **`MaxRetriesExceeded` discarded the root cause.** The last underlying error was only `eprintln!`-ed on intermediate retries, then thrown away. The terminal error/diagnostic said only `"max retries exceeded for {url}"`, with no indication of whether the real cause was DNS failure, a 403, a checksum mismatch, or something else.

### Fix
- Added `DownloadError::HttpStatus { status, url, message, retry_after }` so HTTP response failures carry their status code instead of being flattened into `Network`.
- Added `DownloadError::is_retryable()`: retries connect/timeout errors (`Network`), `Io`, and HTTP `408`/`429`/`5xx`; fails immediately on other 4xx statuses (returns the original error straight to the caller, skipping the retry loop entirely).
- `MaxRetriesExceeded` now carries `attempts: u32` and `source: Box<DownloadError>` (thiserror `#[source]`), and its `#[error(...)]` message interpolates the source so the underlying cause shows up in `Display`/JSON diagnostics without touching any call sites — existing call sites (`wheel_cache.rs`'s `#[from]` conversion, `commands/mod.rs`'s `eprintln!("... {}", e)`) automatically pick up the richer message.
- **Bonus (optional item from the issue):** honors a numeric `Retry-After` header on 429/503 responses for the next backoff delay, falling back to the existing exponential backoff (1s/2s/4s) when absent.

### Judgment calls
- Followed the codebase's existing `thiserror` + `#[source]`/`#[from]` conventions (seen in `build.rs`, `cache.rs`, `lockfile.rs`, etc.) rather than introducing a new error-wrapping pattern.
- Kept `Io` errors retryable (matches prior behavior) since a transient disk/FS hiccup during download setup is plausibly transient; only newly-classified `HttpStatus` errors got fail-fast treatment, since that's what the issue specifically calls out.
- No call sites needed changes: `WheelCacheError::Download` uses `#[from]` (still compiles unchanged), and `commands/mod.rs` prints `DownloadError` via `Display`, which now includes the cause automatically because the format string interpolates `{source}`.
- Did not need a `docs/PLAN.md` update — this issue isn't tracked as one of the PR-A tracks there.

## Test plan
- [x] New unit tests in `src/downloader.rs` (`#[cfg(test)] mod tests`, using `httpmock`):
  - `not_found_fails_fast_without_backoff` — 404 fails in <500ms with exactly 1 request (no retries/backoff).
  - `unauthorized_fails_fast_without_retry` — 401 fails with exactly 1 request.
  - `server_error_is_retried` — 500 is retried up to `max_retries` (3 total requests).
  - `rate_limited_is_retried` — 429 is retried like a 5xx.
  - `max_retries_exceeded_includes_underlying_cause` — the terminal `MaxRetriesExceeded` error's `Display` includes the HTTP status, and `source()` returns the wrapped cause.
  - `successful_download_still_works` — a normal 200 download is unaffected.
- [x] `cargo test` (full suite): **1035 passed, 6 ignored, 0 failed** (52 suites, 167.56s).
- [x] `cargo clippy --all-targets --all-features -- -D warnings`: no issues.
- [x] `cargo fmt -- --check`: clean.

### DIFF
--- a/src/downloader.rs
+++ b/src/downloader.rs
@@ -10,6 +10,10 @@ use thiserror::Error;
 use tokio::fs::{self, File};
 use tokio::io::{AsyncWriteExt, BufWriter};
 
+/// Upper bound on a server-supplied `Retry-After` value, so a malicious or
+/// misconfigured index cannot stall a download indefinitely.
+const MAX_RETRY_AFTER_SECS: u64 = 5;
+
 #[derive(Debug, Clone)]
 pub struct SignatureSpec {
     pub signature: String,
@@ -230,13 +234,15 @@ impl Downloader {
                             source: Box::new(e),
                         });
                     }
-                    // Respect Retry-After on 429/503 when the server sends one;
-                    // otherwise fall back to exponential backoff: 1s, 2s, 4s.
+                    // Respect Retry-After on 429/503 when the server sends one,
+                    // capped to avoid a malicious/misconfigured server stalling
+                    // the download indefinitely; otherwise fall back to
+                    // exponential backoff: 1s, 2s, 4s.
                     let backoff = match &e {
                         DownloadError::HttpStatus {
                             retry_after: Some(secs),
                             ..
-                        } => Duration::from_secs(*secs),
+                        } => Duration::from_secs((*secs).min(MAX_RETRY_AFTER_SECS)),
                         _ => Duration::from_secs(1 << (attempt - 1)),
                     };
                     eprintln!("retrying download {} (attempt {}): {}", url, attempt + 1, e);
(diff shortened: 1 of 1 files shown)


takurot:
Follow-up commit `187e51b`: capped the server-supplied `Retry-After` value (`MAX_RETRY_AFTER_SECS` = 5s) after security review flagged that an unbounded value could stall a download indefinitely (DoS via a malicious/misconfigured index). Added a regression test (`retry_after_is_capped`). Full suite: 1036 passed, 6 ignored; clippy/fmt clean.

Issue #340, opened 2026-07-12
bug(resolver): version ordering is semver-based, not PEP 440 — post-releases, epochs, and numeric pre-release ordering are wrong


takurot:
## Problem

`compare_versions()` / `parse_version_relaxed()` (`src/resolver.rs:1081-1124`) map PEP 440 version strings onto **semver** `semver::Version` semantics. PEP 440 ordering differs from semver in several ways, so version ordering is wrong for whole classes of real PyPI versions.

### 1. Post-releases sort *below* their base release

`parse_version_relaxed("1.0.post1")` → semver `1.0.0-post1`, and semver treats any pre-release suffix as **lower** than the bare version. PEP 440 says `1.0.post1 > 1.0`.

Effect: with candidates `{1.0, 1.0.post1}`, `select_with_constraints()` (`max_by(version_cmp)`) picks `1.0` instead of `1.0.post1`; `pybun outdated` reports a post-release as "older". Packages like `certifi`, `setuptools` and many Debian-style re-uploads use `.postN`.

### 2. Epochs are unparseable → silent fallback to string comparison

`1!2.0` produces suffix `!2.0`, which fails `Version::parse` → `parse_version_relaxed` returns `None` → `compare_versions` falls back to `a.cmp(b)` (byte-wise). So `1!1.0 < 2.0` byte-wise, while PEP 440 mandates the epoch dominates (`1!1.0 > 2.0`).

### 3. Numeric ordering inside pre-release segments is wrong

Semver compares alphanumeric identifiers lexically: `a10 < a2`. PEP 440 compares the numeral numerically: `1.0a10 > 1.0a2`. So for a package publishing `1.0a2 ... 1.0a10`, PyBun selects `1.0a9`/`1.0a2`-era versions as "highest".

### 4. `.devN` releases

`1.0.dev1` becomes semver prerelease `1.0.0-dev1` — accidentally correct direction (below `1.0`), but ordering among `dev`/`a`/`b`/`rc`/`post` segments does not follow PEP 440's `dev < a < b < rc < final < post` chain (e.g. semver sorts `post1 < rc1` lexically? no: `p` > `r` is false — `post` vs `rc`: 'p' < 'r', so `1.0.post1 < 1.0rc1`, while PEP 440 requires `post` > final > `rc`).

## Expected

[...]

PR #353, opened 2026-07-13
fix(resolver): compare ==/!= specifiers via PEP 440 semantics (Issue #339)

### LINKED ISSUE #339: bug(resolver): '==' / '!=' specifiers use raw string equality — violates PEP 440 zero-padding and normalization
## Problem

`VersionSpec::matches()` in `src/resolver.rs` implements `==` and `!=` as **raw string equality**:

```rust
// src/resolver.rs:55,64
VersionSpec::Exact(v) => v == version,
VersionSpec::NotEqual(v) => v != version,
```

All other operators (`>=`, `<`, `~=`, ...) go through `compare_versions()`, but `==`/`!=` do not. PEP 440 requires version comparison for `==` to use **zero-padded release segment comparison** and normalized forms.

## Concrete failures

| Specifier | Candidate on index | PEP 440 | PyBun today |
|-----------|--------------------|---------|-------------|
| `==1.4`   | `1.4.0`            | match   | **no match** → `ResolveError::Missing` |
| `==2024.01` | `2024.1`         | match   | **no match** |
| `!=1.0`   | `1.0.0`            | excluded | **not excluded** → wrong version installed |
| `==1.0.POST1` | `1.0.post1`    | match (case-insensitive normalization) | **no match** |

The `!=` case is the more dangerous direction: a version the user explicitly excluded can still be selected when the index spells it differently (`1.0.0` vs `1.0`).

## Expected

`Exact`/`NotEqual` should compare via the same PEP 440-aware comparison used by the ordered operators (zero-padding release segments, normalizing separators/case), i.e. `compare_versions(version, v) == Ordering::Equal` — or better, a proper PEP 440 version type (see companion issue on semver-based ordering).

## Affected code

- `src/resolver.rs:53-68` (`VersionSpec::matches`)
- Anything flowing through it: `pybun add`, `pybun install`, `pybun upgrade`, lockfile generation, MCP `pybun_resolve`.

## Notes

- Related but distinct from #239 (resolver performance overhaul). If #239 lands a PubGrub/`pep440_rs`-based rewrite this comes for free, but the correctness fix is small and can ship independently.

takurot:
Closes #339

## Summary

`VersionSpec::matches()` implemented `==` / `!=` as raw string equality, violating PEP 440 zero-padding and normalization. The `!=` direction was dangerous: a version the user explicitly excluded (`!=1.0`) could still be selected when the index spells it `1.0.0`.

- `Exact` / `NotEqual` now compare via a new PEP 440-aware `versions_equal()` helper:
  - numeric release-segment comparison with zero padding of the shorter release (`1.4` == `1.4.0`, `2024.01` == `2024.1`, but `1.2.3` != `1.2.3.4`)
  - suffix normalization for case and `-`/`_`/`.` separators (`1.0.POST1` == `1.0.post1`)
  - raw string-equality fallback for unparseable versions
- `compare_versions` / `parse_version_relaxed` are deliberately untouched: routing equality through them would truncate releases to 3 segments (`!=1.2.3` would wrongly exclude `1.2.3.4`) and fail the leading-zero case. Full PEP 440 ordering is tracked separately in #340.

## Test plan

- [x] 9 new unit tests in `src/resolver.rs` (all 4 issue cases + still-matches, not-excluded, string fallback, wildcard non-regression, no-truncation)
- [x] 2 end-to-end tests in `tests/resolver_basic.rs` (`pkg==1.4` resolves against `1.4.0`; `pkg!=1.0` excludes `1.0.0` and selects `0.9.0`)
- [x] TDD: tests written first and observed failing (5 unit + 2 integration failures reproducing the issue verbatim)
- [x] `cargo test` full suite: 1049 passed, 0 failed
- [x] `cargo clippy --all-targets --all-features -- -D warnings`: clean
- [x] `cargo fmt -- --check`: clean

### DIFF
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -52,7 +52,7 @@ impl VersionSpec {
     /// Check whether `version` satisfies this single constraint.
     pub fn matches(&self, version: &str) -> bool {
         match self {
-            VersionSpec::Exact(v) => v == version,
+            VersionSpec::Exact(v) => versions_equal(version, v),
             VersionSpec::Minimum(min) => compare_versions(version, min) != Ordering::Less,
             VersionSpec::MinimumExclusive(min) => {
                 compare_versions(version, min) == Ordering::Greater
@@ -61,7 +61,7 @@ impl VersionSpec {
                 compare_versions(version, max) != Ordering::Greater
             }
             VersionSpec::Maximum(max) => compare_versions(version, max) == Ordering::Less,
-            VersionSpec::NotEqual(v) => v != version,
+            VersionSpec::NotEqual(v) => !versions_equal(version, v),
             VersionSpec::Compatible(base) => is_compatible_release(version, base),
             VersionSpec::Any => true,
         }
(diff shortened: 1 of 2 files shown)


takurot:
## Code Review (rust-review)

**Verdict: APPROVE** — correct, minimal, well-tested fix. No CRITICAL/HIGH findings.

### What was checked
- Both `Exact`/`NotEqual` arms route through the same `versions_equal()` helper — no asymmetry.
- Zero-padding pads the *shorter* release only: `==1.2.3` does not match `1.2.3.4` (regression-tested), avoiding the truncation trap that routing through `parse_version_relaxed` (3-segment semver) would have introduced.
- Leading zeros handled via numeric `u64` parse (`2024.01` == `2024.1`); huge segments overflow the parse → safe fallback to string equality, no panic.
- Suffix normalization (lowercase, strip `./-/_`) fixes `1.0.POST1` == `1.0.post1` without touching ordering semantics (`compare_versions` untouched — #340's territory).
- Ordered operators (`>=`, `<`, `~=`) unaffected; wildcard `==1.4.*` behavior explicitly pinned as unchanged.

### MEDIUM (acceptable as follow-ups, no regression vs. current behavior)
1. **Local version labels**: PEP 440 says a public `==1.0.0` specifier must ignore the candidate's local label, i.e. `==1.0.0` should match `1.0.0+cu118`. `versions_equal` compares the normalized suffix, so it doesn't. Previously (raw string equality) it didn't either — suggest folding into #340's proper version type.
2. **Epoch / `c`↔`rc` aliases**: `1!1.0` vs `1!1.0.0` and `1.0c1` vs `1.0rc1` fall back to normalized-suffix inequality. Same rationale — #340 scope.

### Tests
9 unit + 2 e2e tests, TDD red observed before green. Full suite 1049 passed, clippy/fmt clean, CI 7/7 green. Independently re-verified locally (481 lib tests + 22 resolver_basic + clippy + fmt).

Issue #341, opened 2026-07-12
bug(resolver): pre-release versions are selected by default — PEP 440 requires excluding them unless opted in


takurot:
## Problem

`select_with_constraints()` (`src/resolver.rs:946-953`) picks the highest version among all candidates that satisfy the constraints:

```rust
let candidate = candidates
    .iter()
    .filter(|pkg| constraints.iter().all(|r| r.is_satisfied_by(&pkg.version)))
    .max_by(|a, b| version_cmp(&a.version, &b.version));
```

There is **no pre-release exclusion**. PEP 440 ("Handling of pre-releases") and pip/uv behavior: pre-releases (`a`/`b`/`rc`/`dev`) must be **excluded from version selection by default**, and only considered when (a) the specifier itself mentions a pre-release (`>=2.0rc1`), (b) the user opts in (`--pre`), or (c) only pre-releases exist.

`src/pypi.rs` filters yanked files (`src/pypi.rs:939`) but passes every published version to the resolver as a candidate.

## Concrete failure

If a package's numerically-highest upload is a beta — a routine situation on PyPI (e.g. periods where `numpy 2.0.0b1`, `pandas 3.0.0rc0`, `black 24.x.0a1` are the newest uploads):

```bash
pybun add somepkg        # PyBun: resolves 3.0.0rc0
uv add somepkg / pip     # resolves 2.x.y (latest stable)
```

The agent-first positioning makes this worse: an agent running `pybun add` gets a pre-release pinned into `pybun.lockb` with no warning, and everything downstream (tests, sandbox runs) executes against unstable code.

## Expected

1. Default: filter candidates whose version has a pre/dev segment, unless every candidate is a pre-release or the specifier itself contains one.
2. Add an opt-in flag (`--pre` / `--prerelease=allow`) on `add`/`install`/`upgrade`, mirrored in the MCP `pybun_resolve`/`pybun_install` tools.
3. Emit a diagnostic (`W_PRERELEASE_SELECTED`?) when a pre-release is chosen via the fallback path.

## Affected code

- `src/resolver.rs:946-975` (`select_with_constraints`)
- `src/pypi.rs` candidate construction
- CLI args in `src/cli.rs`, MCP tool schemas in `src/mcp.rs`

## Notes

Depends on being able to *recognize* pre-releases, i.e. the PEP 440 version type from #340. Distinct from #239 (performance).

PR #356, opened 2026-07-15
fix(resolver): honor requires-python metadata when selecting candidates (Issue #342)

### LINKED ISSUE #342: bug(resolver): 'requires-python' metadata from PyPI is ignored — resolves versions incompatible with the active interpreter
## Problem

PyPI publishes a `requires_python` field per release file (JSON API) / `data-requires-python` (simple API). PyBun never reads it: `grep -rn requires_python src/` only hits PEP 723 script metadata parsing (`src/pep723.rs`) — `src/pypi.rs` and `src/resolver.rs` have no trace of it.

So version selection ignores whether a candidate supports the active interpreter. Wheel-tag scoring (`select_artifact`) partially masks this for wheel-only packages, but:

- **sdist fallback**: a version whose wheels don't match can still be selected via sdist, then fail at build/install time with an opaque error instead of resolving an older compatible version.
- **abi3 / py3-none-any wheels**: `py3-none-any` matches any Python 3, so a package that declares `requires-python >= 3.11` still installs on 3.9 — and then fails at *import* time (typically `SyntaxError` on 3.11-only syntax) instead of resolve time.

## Concrete failure

Python 3.9 environment:

```bash
pybun add scipy
# pip/uv: resolve 1.13.x (last 3.9-compatible release)
# pybun: resolves latest (requires-python >=3.10+), py3-only wheel selection may pick
#        an sdist or universal wheel → runtime/build failure
```

For an agent, "resolve succeeded but the script dies on import" is the worst failure mode — the structured diagnostics happen at the wrong layer.

## Expected

1. Capture `requires_python` in `src/pypi.rs` metadata (both JSON and simple API paths) and in the index fixture format (`src/index.rs`).
2. During candidate filtering in the resolver, drop versions whose `requires_python` specifier doesn't match the resolution target Python version (already available — `PYBUN_PYPI_PYTHON_VERSION` / detected interpreter).
3. When all candidates are filtered out, produce a diagnostic that names the Python version conflict (e.g. `E_RESOLVE_PYTHON_INCOMPATIBLE` with hint "newest compatible release is X for Python Y").

## Affected code

- `src/pypi.rs` (release metadata structs, ~line 502 where `yanked` is already parsed)
- `src/resolver.rs` `select_with_constraints`
- `src/index.rs` fixture schema (optional field, backward compatible)

## Notes

The specifier matching needed here ("does 3.9.18 satisfy `>=3.10`") is the same PEP 440 machinery as #339/#340 — one more reason to introduce a proper version/specifier type.

takurot:
Closes #342

## Summary

- Capture the per-release `requires_python` field from the PyPI JSON API (`ReleaseFile` → `CachedPackage` → `ResolvedPackage`) and from index JSON fixtures (`requires_python`, optional and backward-compatible; old binary caches that predate the field are already discarded gracefully by the existing unreadable-cache path).
- `select_with_constraints` now drops candidates whose `requires-python` specifier excludes the resolution target interpreter, so e.g. on Python 3.9 `pybun add scipy` resolves the newest 3.9-compatible release instead of a latest release that fails at build/import time.
- Target version selection: `PYBUN_PYPI_PYTHON_VERSION` override first, then the interpreter detected for the install target. Wired into `install` (detection hoisted before resolution and shared with the Issue #291 cp-tag probe — no extra interpreter spawn), `lock`, `upgrade`, `run` (PEP 723 native path, which already knows its venv's Python), and MCP `pybun_resolve`. `None` (no interpreter found) disables the filter.
- Leniency rule: wildcard (`!=3.0.*`) or unparseable clauses in the metadata are treated as satisfied — imperfect metadata can only widen the candidate set, never wrongly exclude a version.
- When every constraint-matching version is rejected by `requires-python`, resolution fails with a dedicated `ResolveError::PythonIncompatible` (boxed payload to keep the error type small), surfaced as:
  - `E_RESOLVE_PYTHON_INCOMPATIBLE` — names the package, constraint, target Python, and the offending `requires-python` spec
  - `H_RESOLVE_PYTHON_NEWEST_COMPATIBLE` — names the newest release (ignoring constraints) that does support the target Python

## Test plan

- [x] 6 integration tests in `tests/resolver_basic.rs` (filtering, no-target passthrough, missing-metadata compatibility, error payload, newest-compatible hint, leniency)
- [x] 2 unit tests for `requires_python_allows` + 2 index fixture tests
- [x] 2 CLI tests in `tests/cli_install.rs` (skip incompatible / keep newest when compatible)
- [x] 2 JSON diagnostics tests in `tests/json_output.rs` (E_RESOLVE_PYTHON_INCOMPATIBLE, hint)
- [x] 1 MCP test in `tests/mcp.rs` (pybun_resolve filters by PYBUN_PYPI_PYTHON_VERSION)
- [x] Full suite: 1079 passed, 6 ignored (52 suites)

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1509,8 +1509,10 @@ pub(crate) async fn install(
     let working_dir = std::env::current_dir()?;
     let target_env_probe = crate::env::find_python_env(&working_dir)?;
     let python_version_override = python_version_env_override();
-    // PYBUN_FORCE_CP_TAG lets tests (and users) pin the CPython tag deterministically,
-    // bypassing interpreter detection entirely.
+    // PYBUN_FORCE_CP_TAG lets tests (and users) pin the CPython tag deterministically.
+    // Note it no longer bypasses interpreter detection on its own: the detected version
+    // is also needed for `requires-python` filtering, so detection is only skipped when
+    // PYBUN_PYPI_PYTHON_VERSION covers that too.
     let forced_cp_tag = std::env::var("PYBUN_FORCE_CP_TAG")
         .ok()
         .filter(|v| !v.trim().is_empty());
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -18,7 +18,7 @@ use crate::resolver::parse_version_relaxed;
 use crate::resolver::{
     PackageIndex, Requirement, Resolution, ResolveOptions, compare_versions,
     cp_tag_to_dotted_version, current_platform_tags, is_wheel_python_compatible, parse_wheel_tags,
-    python_version_to_cp_tag, resolve, resolve_with_options, select_artifact_for_platform_with_cp,
+    python_version_to_cp_tag, resolve_with_options, select_artifact_for_platform_with_cp,
 };
 use crate::sandbox;
 use crate::sbom;
@@ -1420,6 +1420,29 @@ fn warn_on_prerelease_fallback(resolution: &Resolution, collector: &mut EventCol
     }
 }
 
+/// Explicit `PYBUN_PYPI_PYTHON_VERSION` override for the resolution target
+/// Python version, if set and non-empty.
+fn python_version_env_override() -> Option<String> {
+    std::env::var("PYBUN_PYPI_PYTHON_VERSION")
+        .ok()
+        .map(|v| v.trim().to_string())
+        .filter(|v| !v.is_empty())
+}
+
+/// Python version used for `requires-python` candidate filtering during
+/// resolution (Issue #342): the `PYBUN_PYPI_PYTHON_VERSION` override wins,
+/// otherwise the interpreter detected for the current working directory
+/// (`PYBUN_ENV` / project venv / system Python). `None` — for example when
+/// no interpreter can be found — disables the filter.
+pub(crate) fn resolve_target_python_version() -> Option<String> {
+    if let Some(version) = python_version_env_override() {
+        return Some(version);
+    }
+    let cwd = std::env::current_dir().ok()?;
+    let probe = crate::env::find_python_env(&cwd).ok()?;
+    get_python_version(&probe.python_path).ok()
+}
+
 pub(crate) async fn install(
     args: &crate::cli::InstallArgs,
     collector: &mut EventCollector,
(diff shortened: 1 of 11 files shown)


takurot:
## Code Review (rust-reviewer)

**Verdict: APPROVE** — no CRITICAL/HIGH findings. 1 MEDIUM + LOW items, addressed in `2432fb7` (see below).

### Verified correctness details
- **Pre-release interaction (Issue #341)**: the Python filter runs on constraint-matching candidates *before* the pre-release exclusion/fallback logic, so `PythonIncompatible` fires exactly when constraint-matching versions exist but all are Python-incompatible.
- **Install-path hoisting**: `active_cp_tag` derivation preserves Issue #291 semantics exactly (`PYBUN_FORCE_CP_TAG` short-circuit, `cp311` fallback); no double `get_python_version` spawn. Full `cli_install`/`cli_lock`/`cli_upgrade` suites pass unchanged.
- **bincode cache compatibility**: `requires_python` appended as the last field of `CachedPackage`; old caches fail decode and route through the existing "discarded unreadable PyPI cache entry" self-heal path.
- **JSON fixture / PyPI API**: `#[serde(default)]` keeps old fixtures/responses parsing to `None`.
- **MCP**: `resolve_target_python_version()` computed per-call, not cached at server startup — correct for a long-lived stdio server.
- **`add`/`remove`** inherit the filter via `install()`; no production `resolve()` call site missed.

### Findings & resolution
- **MEDIUM — reconciliation path swallowed `PythonIncompatible`**: the diamond-dependency reconcile branch (`if let Ok(..) else Conflict`) degraded the interpreter conflict into a generic `E_RESOLVE_CONFLICT`, losing the requires-python detail and hint. → **Fixed in `2432fb7`**: the branch now propagates `PythonIncompatible` and keeps `Conflict` for other failures, with regression test `reconciliation_preserves_python_incompatible_error`.
- **LOW — stale `PYBUN_FORCE_CP_TAG` comment** (no longer bypasses detection on its own) → fixed in `2432fb7`.
- **LOW — hint could name a version violating the user's own constraint** → hint now says "(ignoring the version constraint …)" (`2432fb7`).
- **LOW — remaining test-coverage follow-ups** (not blocking): transitive-dependency requires-python E2E, PEP 723 run-path E2E, diagnostic codes not yet listed in docs/PLAN.md.

Post-fix validation: 1080 tests passed (52 suites), clippy `-D warnings` clean, fmt clean.

Issue #294, opened 2026-07-05
bug(run): PEP 723 native installer selects wheels via PATH python despite already knowing the target interpreter version (same root cause as #291)


takurot:
## Summary

When `pybun run` executes a PEP 723 script with the native (non-uv) installer backend, it already resolves and knows the exact target interpreter's version before selecting wheels — but then ignores that value and re-detects the CPython tag from `PATH` instead, via the PATH-based `select_artifact_for_platform`. Same root cause as #291 (fixed for `pybun install` in #292).

## Location

`src/commands/mod.rs`, inside `run_script()`, native PEP 723 install branch:

```rust
// line ~2891-2892: the actual target interpreter is already resolved here
let (base_python, env_source) = find_python_interpreter()?;
let python_version = get_python_version(Path::new(&base_python))?;
...
// line ~3070-3072: but wheel selection ignores `python_version` entirely
for pkg in resolution.packages.values() {
    let selection =
        crate::resolver::select_artifact_for_platform(pkg, &platform_tags);
    ...
}
```

`python_version` is used a few lines later (`major_minor`, for the site-packages path) but is never passed into wheel selection. This is a real execution-time bug (not just a lockfile-metadata issue like #293) — it can download and install a wheel with the wrong CPython ABI into the very venv `base_python`/`python_version` describes.

## Suggested fix

Convert `python_version` to a cp tag with `python_version_to_cp_tag(&python_version)` (already used elsewhere in this file) and pass it to `select_artifact_for_platform_with_cp(pkg, &platform_tags, &active_cp_tag)` instead of the PATH-based call at line ~3072. No new interpreter detection needed — the value is already in scope.

## Related

- #291 (root cause, fixed for `pybun install` in #292)
- #293 (same root cause in `pybun lock`)

PR #358, opened 2026-07-16
fix(test): eliminate process-wide env mutation in parallel tests (Issue #349)

### LINKED ISSUE #349: test: process-wide env::set_var in parallel integration tests is a latent flake — inject config instead
## Summary

Two integration-test files mutate **process-wide** environment variables while the test harness runs other tests in the same binary on parallel threads:

- `tests/pypi_integration.rs:201-233` — sets/restores `PYBUN_PYPI_BASE_URL` and `PYBUN_PYPI_CACHE_DIR` around a `PyPiClient::from_env(...)` call, with a manual save/restore dance
- `tests/cli_run_pep723_native_cp_tag.rs:118-136` — an `EnvGuard` RAII helper doing the same pattern

Both are already wrapped in `unsafe { std::env::set_var(...) }` — Rust 2024 made `set_var` unsafe precisely because of this race. Any test in the same binary that reads those variables (directly, or via `PyPiClient::from_env`, or by spawning the CLI with an inherited environment) can observe the mutated value mid-flight. The save/restore also breaks if two such tests ever interleave. This is a latent flake: it passes today because thread interleavings happen to be favorable, and it will fail rarely and unreproducibly when it doesn't.

## Proposal

In preference order:

1. **Dependency injection over ambient env**: give `PyPiClient` a constructor taking explicit `base_url` / `cache_dir` (keeping `from_env` as the production wrapper that reads env and delegates). The pypi_integration test then never touches the process environment. This also improves the production API (testability without global state).
2. For tests that must exercise the *env-reading* path of the spawned CLI: pass env per-child via `assert_cmd`'s `.env(...)` — child-process env is isolated and race-free; never `std::env::set_var` in the parent.
3. If any in-process env-reading test genuinely remains, serialize it with the `serial_test` crate (`#[serial]`) rather than relying on luck.

## Acceptance

- [ ] Zero `std::env::set_var` / `remove_var` under `tests/` (grep-clean; consider a clippy/CI lint to keep it that way)
- [ ] `PyPiClient` constructible with explicit config in tests

## Suggested area

`tests/pypi_integration.rs`, `tests/cli_run_pep723_native_cp_tag.rs`, `src/pypi.rs` (constructor)

takurot:
Closes #349

## Summary

Two integration-test files mutated **process-wide** environment variables (`unsafe { std::env::set_var }`) while the test harness runs sibling tests of the same binary on parallel threads — a latent flake that passes only while thread interleavings stay favorable.

- **`src/pypi.rs`**: added `PyPiClient::with_config(base_url, cache_dir, offline)` as the primary, dependency-injected constructor. `from_env(offline)` is now a thin production wrapper that reads `PYBUN_PYPI_BASE_URL` / `PYBUN_PYPI_CACHE_DIR` and delegates. This also makes the client testable without global state.
- **`tests/pypi_integration.rs`**: `concurrent_metadata_fetch_is_deduped` builds the client via `with_config` — no env mutation, no save/restore dance.
- **`tests/cli_run_pep723_native_cp_tag.rs`**: this test must exercise the *env-reading* path of `run_script` in-process (the compiled CLI can't host the native PEP 723 installer without a Tokio reactor, see the module docs). The `EnvVarGuard` RAII pattern is replaced with a **child-process re-exec**: the outer test spawns the current test binary selecting an `#[ignore]`d child test, passing all env overrides per-child via `Command::env(...)` (child env is isolated and race-free). The outer test owns the mock PyPI server and still asserts which wheel was downloaded.

## Acceptance criteria (from #349)

- [x] Zero `std::env::set_var` / `remove_var` under `tests/` (grep-clean; remaining hits are doc comments)
- [x] `PyPiClient` constructible with explicit config in tests

## Test plan

- [x] New unit tests: `with_config_uses_explicit_base_and_cache_dir`, `with_config_rejects_invalid_base_url` (TDD: red first)
- [x] `cargo test` — 1102 passed, 7 ignored (the ignored child half is exercised via its outer test)

### DIFF
--- a/src/pypi.rs
+++ b/src/pypi.rs
@@ -153,12 +153,15 @@ fn resolve_pypi_cache_dir(
 }
 
 impl PyPiClient {
-    pub fn from_env(offline: bool) -> Result<Self, PyPiError> {
-        let base =
-            std::env::var("PYBUN_PYPI_BASE_URL").unwrap_or_else(|_| "https://pypi.org".to_string());
-        let normalized = normalize_base(&base)?;
-        let cache_dir_override = std::env::var("PYBUN_PYPI_CACHE_DIR").ok();
-        let cache_dir = resolve_pypi_cache_dir(cache_dir_override.as_deref(), dirs::cache_dir())?;
+    /// Build a client from explicit configuration. This is the primary
+    /// constructor: it takes the base URL and cache directory as arguments so
+    /// callers (and tests) never depend on ambient process environment.
+    pub fn with_config(
+        base_url: &str,
+        cache_dir: PathBuf,
+        offline: bool,
+    ) -> Result<Self, PyPiError> {
+        let normalized = normalize_base(base_url)?;
         Ok(Self {
             base: normalized,
             cache_dir,
@@ -168,6 +171,16 @@ impl PyPiClient {
             deps_once: Arc::new(OnceMap::new()),
             stale_cache_notices: Arc::new(Mutex::new(Vec::new())),
         })
+    }
+
+    /// Production wrapper over [`PyPiClient::with_config`] that reads
+    /// `PYBUN_PYPI_BASE_URL` / `PYBUN_PYPI_CACHE_DIR` from the environment.
+    pub fn from_env(offline: bool) -> Result<Self, PyPiError> {
+        let base =
+            std::env::var("PYBUN_PYPI_BASE_URL").unwrap_or_else(|_| "https://pypi.org".to_string());
+        let cache_dir_override = std::env::var("PYBUN_PYPI_CACHE_DIR").ok();
+        let cache_dir = resolve_pypi_cache_dir(cache_dir_override.as_deref(), dirs::cache_dir())?;
+        Self::with_config(&base, cache_dir, offline)
     }
 
     /// Drain and return any notices about stale/corrupt cache entries that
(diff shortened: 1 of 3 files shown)


takurot:
## Code Review (rust-reviewer subagent)

**Verdict: Approve** — no CRITICAL or HIGH issues. `cargo fmt --check` / `cargo check --tests` / `cargo clippy -D warnings` / both affected test suites verified clean on the diff.

### Confirmed correct
- `PyPiClient::with_config` is a pure function of its arguments (no env reads); `from_env` is a thin delegating wrapper. Clean dependency-injection refactor.
- The re-exec pattern in `cli_run_pep723_native_cp_tag.rs` uses the correct libtest CLI shape (`<name> --exact --ignored`), propagates child panics via exit status with stdout/stderr dumped, and passes env per-child via `Command::env(...)` — race-free by construction.
- Issue #294 regression coverage is preserved: the `fake_wheel_mock.calls() == 1` / `real_wheel_mock.calls() == 0` assertions still run in the outer process against the shared mock server.

### LOW (no action required)
- If the child test is invoked standalone (`--ignored`), the missing `PYBUN_TEST_CP_TAG_SCRIPT` early-return reports "passed" while doing nothing. Mitigated in practice: a silently-skipped child would yield 0 wheel downloads and the outer test's mock-call assertions would fail.

### Follow-up candidates (pre-existing, out of scope)
- `src/env.rs:731` (`test_find_uv_executable_looks_in_path`) and `src/env.rs:553` (`test_pybun_home_override`, `#[ignore]`d) still mutate process-wide env in `--lib` unit tests — same class of race #349 fixes under `tests/`. Worth a follow-up issue applying the same injection / re-exec pattern.

PR #359, opened 2026-07-16
test(resolver): PEP 440 conformance corpus + property tests; fix >/< exclusive comparison (Issue #350)

### LINKED ISSUE #350: test(resolver): add PEP 440 conformance corpus (pypa/packaging) + property-based ordering tests — acceptance suite for #339/#340
## Summary

The resolver's version handling shipped four PEP 440 correctness bugs (#339 `==`/`!=` string equality, #340 semver-based ordering, #341 pre-release selection, #342 requires-python) that `tests/resolver_basic.rs` (582 lines) never caught — because it only tests versions the current implementation happens to handle (plain `X.Y.Z`, simple `rc` suffixes). There is no systematic corpus of the version forms PyPI actually serves (`1.0.post1`, `1!2.0`, `1.0a10`, `1.0.dev1`, `1.0+local`), and no property-based testing anywhere in the repo (no proptest/rstest in Cargo.toml).

This issue adds the **acceptance suite** for the #339/#340 fixes, so the fix PRs land with conformance evidence rather than hand-picked examples.

### 1. Import the pypa/packaging conformance corpus (data-driven test)

`pypa/packaging`'s `tests/test_version.py` maintains the canonical `VERSIONS` list — several hundred version strings **in correct PEP 440 sort order** — plus specifier match/no-match tables in `test_specifiers.py`. Vendor them as a fixture (`tests/fixtures/pep440_versions.txt`, BSD/Apache dual-licensed, attribute in the file header) and assert:

- every adjacent pair compares `Less` (total order matches the reference)
- the specifier table: each `(specifier, version, expected)` triple agrees

Run it against `compare_versions` / `VersionSpec::matches` — today this suite fails (that's the point); it becomes the red test for #339/#340 and stays as the regression net afterwards.

### 2. Property-based invariants (proptest)

Add `proptest` as a dev-dependency with a generator for PEP 440-shaped strings (epoch? release segments, pre/post/dev, local). Invariants:

- ordering is total and antisymmetric: `cmp(a,b) == cmp(b,a).reverse()`
- `==` consistency: `matches(Exact(v), v)` for every generated `v`, including padded forms (`1.0` vs `1.0.0`)
- `!=` is the exact complement of `==`
- normalization idempotence (case, `-/_/.` separators)

### 3. Fixture gap

`tests/fixtures/index_specifiers.json` should gain versions with post/dev/epoch/local segments so the E2E resolve path exercises them, not just the unit comparator.

## Sequencing

Land the corpus test first (marked `#[ignore]` or `#[should_panic]`-style xfail until #339/#340 merge), or land it inside the fix PRs as their test half — either way the corpus is the acceptance criterion.

## Suggested area

`tests/resolver_basic.rs`, `tests/fixtures/`, `Cargo.toml` (dev-dependencies)

takurot:
Closes #350

## Summary

Adds the acceptance suite promised in #350 for the #339/#340 resolver fixes, and fixes the conformance gaps the suite immediately exposed.

### Conformance corpus (vendored from pypa/packaging, commit 9d0ec47)

- `tests/fixtures/pep440_versions.txt` — the canonical 92-entry `VERSIONS` list in reference sort order. `tests/pep440_conformance.rs` asserts every pair agrees with `compare_versions` in both directions (4,186 ordered pairs), plus reflexivity and parseability.
- `tests/fixtures/pep440_specifiers.tsv` — 137 `(specifier, version, expected)` triples from the upstream match table, run through the real `Requirement` parse + `is_satisfied_by` pipeline. Excluded rows are documented in the fixture header: prefix-match specifiers (`==2.*` — intentionally never-matching in PyBun, see `exact_spec_wildcard_behavior_unchanged`) and `===` (unsupported operator). Both fixtures carry upstream attribution (dual Apache-2.0 / BSD-2-Clause).

### Property-based invariants (`proptest`, new dev-dependency)

Structured generator for PEP 440 versions (epoch / release / pre / post / dev / local) plus randomized spellings (case, `-`/`_`/`.` separators, `alpha`/`c`/`rev` aliases, `v` prefix). Invariants: ordering antisymmetry and transitivity, `!=` is the exact complement of `==`, zero-padded release equality (`1.4` == `1.4.0`), normalization equivalence, struct round-trips.

### Fixes the suite caught (first commit)

1. **`>` / `<` exclusive ordered comparison** (19 corpus rows failed): per PEP 440, `>V` must not match post-releases of V's release base (unless V is a post-release) nor local versions of V; `<V` must not match pre-releases of V's release base (unless V is a pre-release). `lib<2.4` now refuses `2.4.0rc1` even with `--pre`, matching pip; `tests/resolver_basic.rs::pre_release_treated_as_less_than_final` was updated accordingly (its old expectation encoded the non-conformant behavior) and a new test covers `--pre` selecting a pre-release *below* the bound.
2. **Parser bug found by the property suite**: `1.0-dev-0` failed to parse — a `-` not followed by digits was treated as a malformed implicit post-release instead of a dev-segment separator. Regression case committed in `tests/pep440_property.proptest-regressions`.

### Fixture gap (issue item 3)

`tests/fixtures/index_specifiers.json` gains post/dev/epoch/local versions; 5 new `cli_install` E2E tests drive them through the full resolve path (post sorts above base; `>` rejects the same-base post; dev excluded without `--pre`; epoch dominates; `==` ignores candidate local labels and prefers the local variant).

## Test plan

- [x] TDD: conformance suite committed red-first against the exclusive-comparison gap, then fixed (Red → Green)
- [x] `cargo test` — 1120 passed (full suite), incl. 4 conformance + 7 property + 5 new E2E tests


takurot:
## Code Review (rust-reviewer subagent)

**Verdict: Request changes** — 1 CRITICAL, 1 MEDIUM, 2 LOW. Diagnostics (fmt/clippy/check/test) all green, but the reviewer verified runtime behavior against a real `pypa/packaging==26.2` install and found correctness gaps in the new exclusive-comparison logic that the vendored corpus does not cover.

### CRITICAL — `base_cmp` is too coarse for the post/pre exclusion rules
`src/resolver.rs:1456` / `:1482` gate exclusion on `base_cmp` (epoch+release only), but pypa/packaging's reference semantics compare against `_post_base(prospective)` (release + **pre**, post/dev/local stripped) for `>`, and `_earliest_prerelease(spec)` (release + pre + **post**, dev pinned to 0) for `<`. Verified-incorrect results vs packaging 26.2:
- `>1.0a1` should match `1.0b1.post1` (packaging: True) — PyBun returns false
- `<2.0.post1` should match `2.0a1` (packaging: True) — PyBun returns false

Neither combination (differing pre-release phase; post of a different pre-release) appears in the vendored table, so the suite passes despite the bug.

### MEDIUM — local-label exclusion in `>` ignores whether the spec itself has a local label
`src/resolver.rs:1460` uses `public_cmp`, which strips local on both sides: `>2.0+abc` wrongly rejects `2.0+xyz`. PEP 440's grammar disallows local on `>`/`<` operands, but PyBun's parser accepts them, so the path is reachable. Needs a `!spec.has_local()` guard.

### LOW
- `base_cmp` naming: once fixed, the two exclusion checks need distinct, more specific helpers rather than one coarse comparison.
- `tests/pep440_property.rs:157` asserts only local-segment *count*, not content.

### Verified correct (no issue)
Parser change (dash-dev) accepts no previously-invalid strings that should stay invalid; `messy()` generator cannot render a different version; `requires_python_allows` inheriting exclusive semantics matches pip behavior (intentional); updated resolver_basic expectations are PEP 440-correct; E2E tests exercise the real resolve→lock path.

Fix for CRITICAL/MEDIUM incoming on this branch with ground-truth-generated regression rows.

takurot:
## Review fixes applied (b2f3eb3)

All CRITICAL/MEDIUM/LOW findings addressed, with pypa/packaging 26.2 as executable ground truth:

- **CRITICAL fixed**: `>V` post-exclusion now keys on the candidate's *post base* (epoch/release/pre-release segment) instead of the coarse `base_cmp`; `<V` exclusion is now "pre-releases at or above V's earliest pre-release (`V.dev0`, local stripped)". Both reviewer cases now agree with packaging (`>1.0a1` matches `1.0b1.post1`; `<2.0.post1` matches `2.0a1`).
- **MEDIUM fixed**: the `>` local-exclusion now requires `!spec.has_local()`, so the lenient-parsed `>2.0+abc` no longer over-excludes `2.0+xyz`.
- **Additional bug found while generating ground truth**: `<=V`/`>=V` must compare the candidate's *public* version — `<=2` now matches `2.0+local` (packaging: True; previous behavior: false).
- **LOW fixed**: property test asserts local segment content (not just count) on round-trip.

Regression net: `tests/fixtures/pep440_specifiers_generated.tsv` — an 836-row cross product (19 specifiers × 44 candidates) generated from `packaging==26.2` (`Specifier(spec, prereleases=True).contains(Version(candidate))`), wired into `pep440_conformance` as a fifth test, plus named unit tests for each reviewer case.

Local verification: full suite 1122 passed, clippy `-D warnings` clean, fmt clean.

PR #365, opened 2026-07-24
fix(profiles): remove dead PYBUN_LOG configuration (Issue #361)

### LINKED ISSUE #361: bug: PYBUN_LOG is set by 'pybun profile' but never read — dead configuration
## Problem

`src/profiles.rs:178` sets the `PYBUN_LOG` environment variable based on the active profile's configured log level (`log_level_str()`), but no code path in the codebase reads `PYBUN_LOG` back to actually control logging output/verbosity.

This is already flagged as a known gap in `CLAUDE.md`:

> Note: `PYBUN_LOG` is currently only *set* by `pybun profile` (to record the profile's configured log level); no code path reads it back to control logging output, so it does not yet do anything if set manually.

## Impact

- Users who set `PYBUN_LOG` manually (expecting standard `env_logger`/`tracing`-style behavior) get no effect — silently ignored, no diagnostic/warning.
- `pybun profile`'s log-level configuration is effectively a no-op today, which is misleading: the profile system implies it controls logging, but it doesn't.

## Suggested Fix

Either:
1. Wire `PYBUN_LOG` into whatever logging/tracing setup pybun uses (e.g. read it during logger initialization in `main.rs`/`entry.rs` and configure the subscriber's filter level accordingly), or
2. If logging control isn't planned soon, remove the `set_var("PYBUN_LOG", ...)` call and the profile's log-level field to avoid the dead/misleading configuration, updating `CLAUDE.md` accordingly.

Option 1 seems preferable since profiles already model dev/prod/benchmark distinctions where differing log verbosity would be genuinely useful.

## Location
- `src/profiles.rs:146` (doc comment describing intent)
- `src/profiles.rs:178` (the `set_var` call)
- `CLAUDE.md:336` (existing note documenting the gap)

takurot:
Closes #361

## Summary
- `pybun profile` set `PYBUN_LOG` (via `ProfileConfig::apply_to_env()`) based on the profile's `log_level`, but no code path anywhere read `PYBUN_LOG` back to control logging output/verbosity — this made the profile system's log-level modeling misleading/dead configuration.
- Removed the `log_level` field, `log_level_str()`, and the `set_var("PYBUN_LOG", ...)` call from `src/profiles.rs`, plus the now-unused `log_level`/`log_level_str` entries from `pybun profile`'s JSON output in `src/commands/tooling.rs`.
- Updated the stale doc note in `CLAUDE.md` and the observability summary in `docs/PLAN.md` to reflect the removal.

## Scope note
`ProfileConfig::apply_to_env()` itself turns out to have zero callers anywhere in the codebase (it only sets `PYBUN_PROFILE`/`PYBUN_TRACE`/custom env vars, none of which anything reads back either). That's a broader dead-code issue outside #361's stated scope (which was specifically about `PYBUN_LOG`), so it's left as-is here — flagging for a possible follow-up issue.

## Test plan
- [x] Added a regression unit test (`profiles::tests::apply_to_env_does_not_set_pybun_log`) verifying `apply_to_env()` no longer sets `PYBUN_LOG` (TDD: confirmed it failed against the old code, then passed after the fix)
- [x] `cargo test` — 1125 passed, 7 ignored, 0 failed
- [x] `cargo fmt` — no diff

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -333,8 +333,6 @@ Diagnostics / dry-run testing:
 - `PYBUN_X_DRY_RUN`: Dry-run override used in tests
 - `PYBUN_X_DRY_RUN_EXIT_CODE`: Simulated exit code override used in tests
 
-Note: `PYBUN_LOG` is currently only *set* by `pybun profile` (to record the profile's configured log level); no code path reads it back to control logging output, so it does not yet do anything if set manually.
-
 ## Common Tasks
 
 ### Adding a New Command
--- a/docs/PLAN.md
+++ b/docs/PLAN.md
@@ -381,8 +381,8 @@ Milestones follow SPECS.md Phase roadmap. PR numbers are suggested grouping; par
   - Risk: 運用/セキュリティ（bind addr, auth）を詰める必要があるので、後回しでもよい。
 - [DONE] PR4.4: Observability layer (structured logging defaults, `PYBUN_TRACE=1` tracing/trace-id propagation, redaction hooks).  
   - Depends on: PR4.1.  
-  - Current: Full observability via `src/schema.rs`. PYBUN_TRACE=1 enables UUID trace IDs. Event streaming with timestamps. Diagnostics array. Schema version tracking. PYBUN_LOG for log level control. Sensitive env vars not leaked in output.
-  - Tests: 9 observability E2E tests (trace_id presence/absence, event timestamps, duration_ms, schema version, diagnostics, env var redaction, log level, event types).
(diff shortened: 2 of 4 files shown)


takurot:
## Code review (self-review via code-reviewer subagent)

**Verdict: APPROVE** — 0 CRITICAL / 0 HIGH / 0 MEDIUM. Build compiles, `cargo clippy --all-features -- -D warnings` clean, all `profiles::` unit tests pass, no documented JSON schema or README contract broken by the removed `log_level`/`log_level_str` fields.

### LOW (non-blocking, follow-up candidates)
1. `src/project.rs:78` has a separate, unrelated `log_level: Option<String>` field on the pyproject.toml-parsing `ProfileConfig` struct (different struct, different module) that also appears to be write-only/dead. Out of scope for #361 — worth its own follow-up issue.
2. `tests/observability.rs::log_level_via_pybun_log` sets `PYBUN_LOG=debug` and only asserts `pybun gc` doesn't crash — it never asserted real behavior even before this PR (vacuous test name/intent). Left as-is since it's still a valid "doesn't crash" smoke test for a user manually setting an unused env var.

No action taken on either LOW item — they're pre-existing and outside this issue's stated scope.

PR #371, opened 2026-07-29
refactor(commands): split install.rs out of commands/mod.rs (Issue #271)

takurot:
## Summary

- Continues the PR-A10 (Issue #186) `src/commands.rs` module-split pattern.
- Extracts `pybun add` / `pybun remove` / `pybun install` / `pybun outdated` / `pybun upgrade` implementations from `src/commands/mod.rs` into a new `src/commands/install.rs`.
- `mod.rs` shrinks from 4903 → 3798 lines; new `install.rs` is 1662 lines.
- No behavior change: JSON detail/diagnostics/summary output and CLI dispatch are unchanged (`mod.rs` now calls `install::run_add(...)`, `install::run_remove(...)`, etc.).
- This PR covers only the `install.rs` portion of Issue #271's proposed scope. `lock.rs` / `run.rs` / `build.rs` / `x.rs` / `python_env.rs` / `project.rs` extraction remain as follow-up work (tracked in `docs/PLAN.md` PR-A10 `Remaining:` note).

Progress on #271 — not closing yet, since #271 also proposes `run.rs` / `lock.rs` extraction which is out of scope for this PR.

## Test plan

- [x] `cargo test --test '*'` — passes (pre-existing `cli_run` failures are caused by a broken local Homebrew `python@3.14` install unrelated to this change; confirmed identical failures reproduce on `main` via `git stash`)
- [x] `cargo test --test e2e_general` — 9 passed
- [x] `just lint` (clippy + fmt check) — clean, zero warnings
- [x] `cargo fmt -- --check` — clean
- [x] `docs/PLAN.md` PR-A10 entry updated with this extraction's summary and test results

Issue #344, opened 2026-07-12
refactor(mcp): decompose src/mcp.rs (3292 lines) into transport / schema / per-tool modules


takurot:
## Summary

`src/mcp.rs` is now the largest file in the codebase after `commands/mod.rs`: **3292 lines** in a single module mixing four unrelated concerns:

1. JSON-RPC 2.0 framing / stdio transport loop
2. Tool schema definitions — `handle_tools_list` is a **318-line function** (`src/mcp.rs:1031`) of hand-written inline `serde_json::json!` schema blobs
3. Per-tool handlers (`call_run`, `call_install`, `call_resolve`, `call_gc`, `call_doctor`, ...)
4. Audit logging and sandbox glue

## Proposal

Split following the same pattern already applied to `commands/` (#186, #271):

```
src/mcp/
├── mod.rs        # public API, server startup (stdio loop)
├── rpc.rs        # JSON-RPC 2.0 request/response framing, error codes
├── schema.rs     # tool definitions as data (typed structs → serialized), not a 318-line json! block
└── tools/
    ├── mod.rs    # dispatch
    ├── run.rs
    ├── install.rs
    ├── resolve.rs
    └── maintenance.rs  # gc / doctor / audit
```

Making tool definitions data-driven (a `ToolDef { name, description, input_schema }` table) also removes the risk of `tools/list` drifting from what `call_tool` actually accepts — today the schema and the validation logic live ~1000 lines apart.

## Sequencing

Best done **after** #272 (MCP handlers delegate to shared command-layer implementations): once handlers shrink to thin adapters, the per-tool modules become trivial. Doing the file split first would just move soon-to-be-deleted duplicated logic around.

Part of the phased refactoring plan (see tracking issue).

## Suggested area

`src/mcp.rs`

takurot:
Current `origin/main` was reconciled against this Issue, and the requested decomposition is already implemented by merged PR #373.

Evidence:
- the former 3,324-line `src/mcp.rs` was removed;
- transport/server state now lives in `src/mcp/mod.rs`;
- tool definitions moved to `src/mcp/schema.rs` as a `Vec<Tool>` data table;
- audit and sandbox glue moved to `src/mcp/audit.rs`;
- handlers moved under `src/mcp/tools/{run,install,resolve,maintenance}.rs`;
[...]

PR #373, opened 2026-07-30
refactor(mcp): decompose src/mcp.rs into transport/schema/audit/tools modules (Issue #344)

takurot:
## Summary

Splits the 3324-line `src/mcp.rs` into a directory module, following the pattern established in PRs #371/#372 for `commands/`:

- `mcp/mod.rs` — JSON-RPC transport (`JsonRpcRequest`/`JsonRpcResponse`), `McpServer` struct, request dispatch, stdio server loop
- `mcp/schema.rs` — `tools/list` tool definitions, previously an inline 318-line `json!` block in a single method, now `build_tools_list()`
- `mcp/audit.rs` — audit log recording, file-write snapshotting, sandbox-policy helpers, ruff-tool-runner glue
- `mcp/tools/` — per-tool handlers grouped by concern: `resolve.rs`, `install.rs`, `run.rs`, `maintenance.rs` (gc/doctor/lint/type_check/profile/fix/context/drift/test/audit/upgrade)

All 14 `call_*`/`read_*` tool methods on `McpServer` never referenced `self`, so they were converted to free functions (grouped by concern) rather than requiring struct-field visibility workarounds.

**Deviation from Issue #344's literal proposed layout:** no separate `rpc.rs`. JSON-RPC framing (`handle_request`, `handle_initialize`, `handle_tools_call`, etc.) stays in `mod.rs` alongside `McpServer`, since these methods are tightly coupled to `&mut self` state (`initialized`, `session_id`, `audit_log`) and splitting them out would leave two halves with no clean seam. This PR addresses the schema-inlining and giant-`impl`-block pain points the issue was primarily about.

Zero behavior changes — mechanical move/regroup, no logic altered.

## Test plan

- [x] `cargo build --lib` — compiles clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — no issues
- [x] `cargo fmt` — applied
- [x] `cargo test --test mcp` — 64 passed, 4 failed (identical 64/4 split reproduced on `main` before this change — local dyld/venv Python subprocess issue, unrelated to this refactor)
- [x] `cargo test --test json_schema` — 23 passed
- [x] `cargo test --test e2e_general` — 9 passed
- [x] `cargo test --test cli_smoke` — 6 passed
- [x] `cargo test --test resolver_basic` — 37 passed
- [x] `cargo test --lib` — all unit tests (incl. relocated `mcp::tests`) pass

Progress on #344.


takurot:
## Code Review — Issue #344 mcp.rs decomposition

**Verdict: Approve**

### Verification performed
- Extracted `main:src/mcp.rs` (pre-refactor) and diffed all 16 tool-handler function bodies (`call_resolve`, `call_install`, `call_run`, `call_gc`, `call_doctor`, `call_lint`, `call_type_check`, `call_profile`, `call_fix`, `read_cache_info`, `read_env_info`, `call_context`, `call_drift`, `call_test`, `call_audit`, `call_upgrade`) against their new locations under `src/mcp/tools/`, normalizing only for signature form (`&self` removal, visibility keyword). **Zero body-content mismatches** — every function is logically identical to the original.
- Confirmed the `handle_tools_call` dispatch match arms in `src/mcp/mod.rs` map to the exact same 14 tool names → `tools::call_*`/`tools::read_*` free functions as the original `self.call_*`/`self.read_*` methods, in the same order.
- Confirmed `audit::snapshot_for_tool` / `audit::diff_file_writes` / `audit::build_audit_entry` call sites in `mod.rs` correspond 1:1 with the original inline calls.
- Confirmed clean `cargo build --lib`, `cargo fmt`, and `cargo clippy --all-targets --all-features -- -D warnings` (0 warnings) per the PR description.
- Confirmed the 64 passed / 4 failed `tests/mcp.rs` result is identical on unmodified `main` (verified via `git stash` in the prior session) — the 4 failures are a pre-existing local dyld/venv environment issue unrelated to this change, not a regression.

### Findings

**MEDIUM**
- `src/mcp/tools/maintenance.rs` (1275 lines, 11 handlers: gc/doctor/lint/type_check/profile/fix/context/drift/test/audit/upgrade) is a fairly large grab-bag file, well over the project's 800-line guideline in CLAUDE.md. It's a reasonable interim grouping ("everything that isn't resolve/install/run") and each function is independently comprehensible, so I don't consider this a blocker, but a natural follow-up would be to split it further (e.g. `quality.rs` for lint/type_check/fix, `diagnostics.rs` for doctor/context/drift, `test.rs`/`upgrade.rs` standalone) — worth a follow-up issue rather than blocking this PR.

**LOW**
- `src/mcp/tools/*.rs` functions and the `tools::*` re-exports in `tools/mod.rs` are declared `pub(crate)`, but they're only ever consumed from `mcp::mod.rs`. `pub(super)` would express the actual visibility need more precisely and matches the pattern already used in `audit.rs`/`schema.rs`. Not a correctness issue, just a minor hygiene nit.

**Non-blocking note**
- The deviation from Issue #344's originally proposed layout (no separate `rpc.rs`; JSON-RPC transport stays in `mod.rs`) is well-justified and clearly documented in the PR body — agreed this is the right call given the tight coupling to `&mut self` state.

No CRITICAL or HIGH issues found. CI is green (8/8 checks). Good to merge.

PR #388, opened 2026-08-01
fix(mcp): gate unsafe_no_sandbox behind server opt-in + prevent script path traversal (Issue #377)

### LINKED ISSUE #377: MCP run tool: unauthenticated sandbox disable + path traversal
## Summary
`src/mcp/tools/run.rs` allows any MCP client to fully disable the sandbox and to escape the intended script directory.

## Details

- The `unsafe_no_sandbox` flag is read directly from caller-provided MCP tool arguments with **zero authorization or allowlist check**. Any MCP client can pass `{"unsafe_no_sandbox": true}` to execute arbitrary code outside the sandbox.
- The `script` path has no canonicalization or path-traversal prevention — `../../etc/cron.daily/exploit` is accepted as-is.

## Suggested Fix
- Gate `unsafe_no_sandbox` behind a server-side allowlist/config flag that the MCP caller cannot set directly (e.g. must be enabled via `pybun mcp serve --allow-unsafe-no-sandbox` or similar operator-controlled config, not client-supplied JSON).
- Canonicalize the `script` path and verify it resolves within the intended project/script root before executing; reject with a diagnostic if it escapes.

## Location
`src/mcp/tools/run.rs`

## Related
See also #376 (sandbox isolation bypasses) — this is the entry point that lets a caller skip the sandbox entirely.

takurot:
Closes #377

## Summary
- `pybun_run`'s `unsafe_no_sandbox: true` argument previously disabled the MCP sandbox unconditionally whenever a client requested it — any MCP client could remotely bypass the sandbox with no server-side authorization.
- `unsafe_no_sandbox` is now only honored if the server operator explicitly started `pybun mcp serve` with a new `--allow-unsafe-no-sandbox` flag (wired through `PYBUN_MCP_ALLOW_UNSAFE_NO_SANDBOX`). If the client requests it but the server wasn't opted in, the request is denied, the sandbox stays enabled, and a new `W_MCP_UNSAFE_NO_SANDBOX_DENIED` warning is returned.
- The `script` argument to `pybun_run` is now canonicalized and checked for containment within the current working directory, rejecting path traversal attempts (e.g. absolute/relative paths resolving outside the project root).
- Updated the `pybun_run` tool schema description for `unsafe_no_sandbox` to document the new gating requirement.

## Test plan
- [x] `cargo build`
- [x] `cargo fmt -- --check`
- [x] New test: `mcp_pybun_run_unsafe_no_sandbox_denied_without_server_gate` — asserts default-deny behavior
- [x] Updated test: `mcp_pybun_run_unsafe_no_sandbox_allows_explicit_opt_out_with_warning` — now passes the gating env var explicitly
- [x] New test: `mcp_pybun_run_script_path_traversal_rejected` — asserts a script path outside the project directory is rejected
- [x] `cargo test --test mcp` (18 mcp tests pass; 4 unrelated pre-existing failures reproduce identically on `main` due to a broken local Python 3.14 install, confirmed via `git stash`)

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -411,6 +411,11 @@ pub struct McpServeArgs {
     /// Use stdio mode for MCP communication.
     #[arg(long)]
     pub stdio: bool,
+    /// Allow clients to set `unsafe_no_sandbox: true` on `pybun_run` calls.
+    /// Disabled by default; client-supplied JSON-RPC arguments alone can
+    /// never disable the sandbox (Issue #377).
+    #[arg(long)]
+    pub allow_unsafe_no_sandbox: bool,
 }
 
 #[derive(Subcommand, Debug)]
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -643,6 +643,13 @@ pub async fn execute(cli: Cli) -> Result<()> {
         }
         Commands::Mcp(cmd) => match cmd {
             McpCommands::Serve(args) => {
+                if args.allow_unsafe_no_sandbox {
+                    // SAFETY: single-threaded at startup, before any concurrent
+                    // access to the environment begins.
+                    unsafe {
+                        std::env::set_var("PYBUN_MCP_ALLOW_UNSAFE_NO_SANDBOX", "1");
+                    }
+                }
                 if args.stdio {
                     // Run MCP server in stdio mode - this blocks until shutdown
                     if let Err(e) = crate::mcp::run_stdio_server().await {
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -238,6 +238,7 @@ mod tests {
             command: Commands::Mcp(McpCommands::Serve(McpServeArgs {
                 port: 9999,
                 stdio: true,
+                allow_unsafe_no_sandbox: false,
             })),
         };
         assert!(requires_tokio_runtime(&cli));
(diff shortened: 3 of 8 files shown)


takurot:
Ran an independent security review of this PR before merge.

**Findings:**
- ✅ The `unsafe_no_sandbox` gating via `PYBUN_MCP_ALLOW_UNSAFE_NO_SANDBOX` / `--allow-unsafe-no-sandbox` is sound — it's set once at server startup and is never reachable from client-supplied JSON-RPC arguments (including via `sandbox_policy.allow_env`, which only whitelists env vars passed to the sandboxed *child process*, not the server's own environment).
- ⚠️ HIGH: `call_run` canonicalized `script_path` solely to validate path containment against `cwd`, but then passed the original, unresolved string to `RunArgs::target` for actual execution. This left a TOCTOU (time-of-check to time-of-use) gap: a symlink swapped in between the check and execution could bypass the containment guarantee.

**Fix (f2a0547):** the canonicalized path is now threaded through to every downstream consumer — dry-run description/risk estimation and the real execution target — instead of being discarded after validation. Verified with `cargo build`, `cargo clippy --all-targets --all-features -- -D warnings` (clean), `cargo fmt -- --check` (clean), and `cargo test` (513+ passed; the only failures are the 28 pre-existing `tests/cli_run.rs` failures caused by a broken local Python 3.14 install, unrelated to this change).

PR #391, opened 2026-08-02
fix(pypi): correct PEP 440/508 marker comparison and error propagation (Issue #381)

### LINKED ISSUE #381: PEP 440/508 version and marker comparison logic is incorrect in src/pypi.rs
## Summary
Version comparison and environment-marker evaluation in `src/pypi.rs` have correctness bugs that can silently corrupt dependency resolution.

## Details

- `compare_versions` uses exact tuple equality for `==`, so `python_version == "3.10"` returns false on Python 3.10.1. Per PEP 440, this should be a prefix match against the major.minor version.
- `eval_python_version_marker` uses a substring check that also matches `python_full_version`, a different PEP 508 marker variable — causing markers to be evaluated against the wrong variable.
- `fetch_requires_dist` returns an **empty dependency list on ANY non-success HTTP status** (500, 503, network error), silently producing packages with zero dependencies and corrupting resolution instead of surfacing an error/retry.

## Suggested Fix
- Implement `python_version == "X.Y"` as a major.minor prefix match per PEP 440, not exact tuple equality.
- Parse PEP 508 marker variable names precisely (e.g. tokenize instead of substring-matching) so `python_version` and `python_full_version` are never confused.
- On non-success HTTP responses in `fetch_requires_dist`, propagate an error/diagnostic (or retry) instead of silently returning an empty dependency list.

## Location
`src/pypi.rs`

takurot:
## Summary
- `python_version == "X.Y"` markers now match a major.minor prefix per PEP 440, instead of requiring exact major.minor.patch tuple equality (so `python_version == "3.10"` correctly matches `3.10.7`).
- `python_version` and `python_full_version` marker variables are now distinguished by tokenizing the identifier immediately preceding the comparison operator, instead of substring-matching — previously a `python_full_version` marker could be mistakenly evaluated as `python_version`, and compound markers like `os_name == "nt" and python_version >= "3.8"` could match the wrong variable.
- `fetch_requires_dist` now propagates non-success HTTP responses (e.g. 500/503) as a `PyPiError::Http` error instead of silently returning an empty dependency list, which previously could corrupt dependency resolution.

Closes #381

## Test plan
- [x] Added unit tests in `src/pypi.rs` covering major.minor prefix matching, `python_version`/`python_full_version` disambiguation, and marker-variable tokenization in compound markers
- [x] Added integration test `version_metadata_server_error_surfaces_as_error` in `tests/pypi_integration.rs` covering HTTP 500 propagation
- [x] `cargo test --lib` — all pass (22/22 new + existing)
- [x] `cargo test --test '*'` — all suites pass except `cli_run` (pre-existing failures unrelated to this change, caused by a broken local Python 3.14 framework install: `dyld: Library not loaded`)
- [x] `cargo test --test e2e_general` — 9/9 pass
- [x] `just lint` (clippy `-D warnings`) — clean

### DIFF
--- a/src/pypi.rs
+++ b/src/pypi.rs
@@ -315,7 +315,9 @@ impl PyPiClient {
             .map_err(|e| PyPiError::Parse(e.to_string()))?;
         let resp = self.http.get(url).send().await?;
         if !resp.status().is_success() {
-            return Ok(Vec::new());
+            return Err(PyPiError::Http(
+                resp.error_for_status().unwrap_err().to_string(),
+            ));
         }
         let body: VersionResponse = resp.json().await?;
         Ok(body.info.requires_dist.unwrap_or_default())
@@ -675,10 +677,9 @@ pub(crate) fn marker_allows(marker: &str, py_version: &str) -> bool {
         return false;
     }
 
-    // Handle simple python_version comparisons; if parsing fails, allow by default
-    if marker.contains("python_version")
-        && matches!(eval_python_version_marker(&marker, py_version), Some(false))
-    {
+    // Handle simple python_version/python_full_version comparisons; if
+    // parsing fails, allow by default
+    if matches!(eval_python_version_marker(&marker, py_version), Some(false)) {
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


takurot:
## Sub-agent code review (rust-reviewer)

**Scope:** diff-only review of `src/pypi.rs` and `tests/pypi_integration.rs`.

### Summary
The three fixes are correct and well-targeted. No CRITICAL or HIGH issues found.

### Findings

| Severity | Location | Finding |
|---|---|---|
| MEDIUM | `eval_python_version_marker` | `major_minor()` truncates only the runtime version (lhs), not the marker's rhs. A non-standard 3-component rhs (e.g. `python_version == "3.10.7"`) would compare `(3,10,0)` against `(3,10,7)`, which can never satisfy `==` and can flip `>=`/`<` results. Will fix by truncating rhs to major.minor as well when the marker variable is `python_version`. |
| MEDIUM | `fetch_requires_dist` | `resp.error_for_status().unwrap_err()` after `!resp.status().is_success()` would panic if status were in the 1xx/3xx range (only guaranteed safe for 4xx/5xx). Low probability in practice (reqwest follows redirects), and it mirrors an existing pattern elsewhere in the file, but will switch to using `resp.status()` directly to avoid the `unwrap_err()` panic risk entirely. |
| LOW | `lhs_var` extraction | Only supports the "variable on the left" marker form (`python_version >= "3.8"`), not the reversed form. Pre-existing scope, not a regression from this diff — leaving as-is. |
| LOW | `eval_python_version_marker` | Only evaluates the first recognized comparison in a marker string; compound markers combining two `python_version` clauses via `and` only honor the first. Pre-existing limitation, unrelated to the bugs reported in #381 — leaving as-is, worth a follow-up issue if it matters in practice. |

Addressing the two MEDIUM findings now since they're directly in the code this PR touches.

PR #394, opened 2026-08-04
fix(install): close supply-chain signature-verification gaps (Issue #378)

### LINKED ISSUE #378: Install scripts (install.sh / install.ps1) provide false sense of supply-chain security
## Summary
Both `scripts/install.sh` and `scripts/install.ps1` have systemic signature-verification weaknesses that undermine the supply-chain integrity they're meant to provide.

## Details

- **Self-signed manifest**: The signature value AND the public key are read from the *same* downloaded manifest. An attacker who controls the manifest controls both — verification proves nothing.
- **Partial-manifest bypass**: If only one of `$SigValue`/`$SigPub` is present, verification is silently skipped with no warning. A tampered manifest that strips the signature fields falls back to checksum-only mode.
- **HTTP accepted**: `scripts/install.sh` accepts `http://*` URLs alongside `https://*` for manifest download, enabling trivial MITM.
- **`eval` on external data**: Parsed manifest fields are passed to `eval` in `install.sh`, creating a code-injection surface.
- **Asset URL override**: A manifest can redirect downloads to arbitrary schemes/hosts.

## Suggested Fix
- Embed the trusted public key in the install script itself (or fetch it from a separate, pinned, out-of-band location) — never trust a key delivered alongside the thing it's meant to verify.
- Fail closed (abort install) if signature fields are missing or malformed, rather than silently degrading to checksum-only.
- Reject non-HTTPS manifest/asset URLs outright.
- Replace `eval` on parsed manifest fields with safe field access (no shell evaluation of external data).
- Validate asset URL scheme/host against an allowlist before downloading.

## Location
`scripts/install.sh`, `scripts/install.ps1`

takurot:
Closes #378

## Summary

`scripts/install.sh` and `scripts/install.ps1` had five systemic signature-verification weaknesses that undermined the supply-chain integrity they were meant to provide. This PR fixes all of them:

- **Self-signed manifest**: The trusted minisign public key is now hardcoded in both install scripts (`TRUSTED_MINISIGN_PUBLIC_KEY` / `$TrustedMinisignPublicKey`), matching `security/pybun-release.pub`. The manifest's own `signature.public_key` field is parsed but never used for verification.
- **Partial-manifest bypass**: Verification now fails closed — if `SIG_VALUE` is empty, install aborts with "manifest missing signature value for asset (refusing to install unsigned artifact)" instead of silently degrading to checksum-only mode.
- **HTTP accepted**: Non-HTTPS manifest sources (`http://*`) are now rejected outright in both scripts; only `https://` and `file://` are accepted.
- **`eval` on external data**: Manifest field parsing no longer uses `eval`. Python emits `NAME=VALUE` lines with backslash/newline/CR escaping; the shell reads them via `while IFS= read -r line < file` and decodes with `printf '%b'` (a fixed literal format string, so no injection risk regardless of field content).
- **Asset URL override**: Manifest-supplied asset URLs are now validated against an HTTPS + host allowlist (`ASSET_URL_HOST_ALLOWLIST`, default `github.com`) via `validate_asset_url()` / `Test-AssetUrl` before being accepted.

New env var overrides (for testing/ops):
- `PYBUN_INSTALL_TRUSTED_PUBKEY` — override the trusted pubkey
- `PYBUN_INSTALL_ASSET_HOST_ALLOWLIST` — override the allowlist
- `PYBUN_INSTALL_ASSET_ALLOW_INSECURE` — narrow, default-off escape hatch permitting `http://` asset URLs against an allowlisted host, needed because `httpmock` (our test HTTP mock server) has no TLS support. Never set in production.

## Test plan

- [x] New tests in `tests/install_scripts.rs`: HTTP manifest source rejection, out-of-allowlist asset URL rejection, manifest field injection inertness (proves `eval` removal via canary-file check), ignored manifest-supplied public key, fail-closed install when signature missing, full signed install using a real `minisign`-generated keypair via `httpmock::MockServer`
- [x] Updated pre-existing `tests/docs.rs` installer test fixture to use an allowlisted host instead of `example.com`
- [x] `cargo test --test '*'`: 627 passed, 5 ignored, no regressions
- [x] `cargo clippy --all-targets --all-features -- -D warnings`: no issues
- [x] `cargo fmt -- --check`: clean
- [x] `sh -n scripts/install.sh`: syntax OK

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -20,6 +20,43 @@ $IsMacOS = [System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform(
 $AliasName = "pybun-cli"
 $AliasBinaryName = if ($IsWindows) { "pybun-cli.exe" } else { "pybun-cli" }
 
+# Trusted minisign public key for verifying release artifacts (matches
+# security/pybun-release.pub). Baked into the script so a compromised/MITM'd
+# manifest cannot supply its own key and self-sign a malicious artifact.
+$TrustedMinisignPublicKey = if ($env:PYBUN_INSTALL_TRUSTED_PUBKEY) {
+    $env:PYBUN_INSTALL_TRUSTED_PUBKEY
+} else {
+    "RWQtfZNEyNnATcgzxfwDB+iNwGwW8bwflfnjHMizq6H84B1VRXO92yZ5"
+}
+
+# Hosts allowed for asset download URLs (comma separated).
+$AssetUrlHostAllowlist = if ($env:PYBUN_INSTALL_ASSET_HOST_ALLOWLIST) {
+    $env:PYBUN_INSTALL_ASSET_HOST_ALLOWLIST
+} else {
+    "github.com"
+}
+
+# Test-only escape hatch to allow http:// asset URLs against an allowlisted
+# host (e.g. a local mock server in integration tests). Defaults to disabled;
+# never set this in production use.
+$AssetUrlAllowInsecure = $env:PYBUN_INSTALL_ASSET_ALLOW_INSECURE -eq "1"
+
+function Test-AssetUrl {
+    param([string]$Url)
+    if ($Url -match "^https://") {
+        # ok
+    } elseif ($Url -match "^http://" -and $AssetUrlAllowInsecure) {
+        # ok (test-only)
+    } else {
+        throw "insecure asset URL rejected (must be https): $Url"
+    }
+    $uri = [System.Uri]$Url
+    $allowed = $AssetUrlHostAllowlist -split "," | ForEach-Object { $_.Trim() }
+    if ($allowed -notcontains $uri.Host) {
+        throw "asset URL host not in allowlist ($AssetUrlHostAllowlist): $($uri.Host)"
+    }
+}
+
 function Write-Log {
     param([string]$Message)
     if ($Format -eq "json") {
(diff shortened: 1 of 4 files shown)


takurot:
## Code review (security-reviewer subagent)

Ran an independent security review of this PR's diff before merge.

### HIGH - fixed in c4eeb83
`validate_asset_url()` in `scripts/install.sh` extracted the URL host via naive shell string manipulation (`${host%%/*}`, `${host%%:*}`), which does not handle userinfo syntax (`user:pass@host`) per RFC 3986. A manifest-supplied asset URL like `https://github.com:@evil.com/pybun-asset.tar.gz` mis-parsed to `host=github.com` (passing the allowlist check), while `curl`/`wget` given the raw URL actually connect to `evil.com` -- fully defeating the host allowlist this PR introduces. `install.ps1`'s `Test-AssetUrl` was unaffected (it already used `[System.Uri]$Url` -> `.Host`, which resolves userinfo correctly).

**Fix**: `validate_asset_url()` now delegates host/scheme extraction to Python's `urllib.parse.urlsplit()` (python3 is already a hard dependency of the script), matching the approach already used in `install.ps1`. Added regression test `install_sh_rejects_asset_url_with_userinfo_host_confusion` covering the exact bypass URL.

### MEDIUM/LOW - informational, not blocking, considered out of scope for this PR
- `ASSET_NAME_FROM_MANIFEST`/`$asset.name` is used unsanitized in local filesystem paths (`$tmp_dir/$ASSET_NAME`); a `../` traversal name could theoretically write outside the temp dir. Impact is bounded since the downloaded content still must pass sha256 + signature verification before use. Pre-existing behavior, not introduced by this PR.
- Embedded NUL bytes in a manifest JSON string could cause silent truncation in the `emit()`/`printf '%b'` escape/decode scheme. Not exploitable for injection (the format string itself is fixed, not attacker-controlled).
- `$(...)` command substitution strips trailing newlines when decoding field values via `printf '%b'`. Currently a no-op since signature content is already stripped, but a latent fragility if a future field needs trailing-newline preservation.
- The `ASSET_URL_ALLOW_INSECURE` test-only escape hatch (default off, needed because `httpmock` 0.8.3 lacks TLS support) is correctly scoped -- still subject to the (now-fixed) allowlist. Reviewer suggested optionally hard-restricting the insecure-http path to loopback hosts only as a future hardening, non-blocking.

No issues found with: the eval-free `printf '%b'` field decode (fixed literal format string, not derived from untrusted input -- no format-string injection), the escape ordering in `emit()` (prevents raw newlines from smuggling fake `NAME=VALUE` lines), or the field-name whitelisting via the shell `case` statement.

All fixes verified: `cargo fmt -- --check`, `cargo test --test '*'` (628 passed, 0 regressions), `cargo clippy --all-targets --all-features -- -D warnings` (clean).

